PR #15, opened 2018-03-21
add modemmanager release candidate

ryantm:
https://lists.freedesktop.org/archives/modemmanager-devel/2018-January/006157.html announces the release candidate


AMDmi3:
I don't particularly like the discrepancy between `1.7.990` and `1.8-rc1`. Perhaps it'd be better to use `setver` to just convert `1.7.990` to `1.8-rc1`. It'll be devel by default.

Also, for future, '[0-9]+' should have been used here instead '[0-9]*' as we intend to require a version component, and it can't be empty.

ryantm:
Updated commit based on feedback.

PR #17, opened 2018-03-24
add wayland alpha and beta

ryantm:
https://wayland.freedesktop.org/releases.html


ryantm:
I did it this way because it seems similar to what you suggested I do in https://github.com/repology/repology-rules/pull/15/files . They are calling the releases alpha and beta in their release, so they should be devel, right?

AMDmi3:
In #15 there was weird dual versioning scheme. Nothing like that here.

PR #18, opened 2018-03-25
make wayland and weston alpha and beta releases devel

ryantm:
https://wayland.freedesktop.org/releases.html says these versions are alpha and beta releases


AMDmi3:
Please use generic rules with regexps, e.g. `.*\\.9[0-9]`

ryantm:
Thanks for the feedback; I updated the PR.

PR #24, opened 2018-05-02
Merge apache-arrow into apache-cpp separate from python:pyarrow

veprbl:
Techicaly apache-cpp and pyarrow are two distinct packages even though they are
built from the same tarball (and there is also a java version). apache-cpp is a
dependency of pyarrow. There is also similarly named relevant package
parquet-cpp.

### DIFF
--- a/800.renames-and-merges/a.yaml
+++ b/800.renames-and-merges/a.yaml
@@ -86,8 +86,6 @@
 - { setname: apache,                   name: apache-tools, addflavor: tools }
 - { setname: apache,                   name: htdigest, addflavor: htdigest }
 - { setname: apache,                   namepat: "apache[0-9.]+" }
-- { setname: apache-arrow,             name: "python:pyarrow" }
-- { setname: apache-arrow,             name: apache-arrow-glib, addflavor: glib }
 - { setname: apcupsd,                  name: [apcupsd-x11,apcupsd-cgi] }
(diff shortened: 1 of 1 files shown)


AMDmi3:
I don't see why they should be split. Merging arrow-cpp into apache-arrow is good though.

veprbl:
I'm not convinced myself, but I will try to explain. My argument is that ```arrow-cpp```, ```pyarrow``` (as well as java, js, rust, and so on) have distinct build processes and as such will likely be packaged as separate packages (and that is the case now). And then it would be less confusing to see which packages are shipped by which distribution. Say, Homebrew will not package ```pyarrow```, but when the list [1] combines c++ and python implementations this is not so apparent to the reader anymore.

[1] https://repology.org/metapackage/apache-arrow/versions

AMDmi3:
Technically you are right. However, repology focuses more on _projects_ than individual packages, and a single distfile is most definitely a single project. Besides, keeping such project as a set of separate packages in general is not that straightforward, as some repositories may package everything as a single package, and we don't have a way to split it (and don't want to). Or there's a single source packages for a separated binary ones, with the same problem. So unless there's a really good reason, I'd prefer to leave this merged.

veprbl:
You have a point. Let's consider this resolved by 3b75af07581d4c761c62d237d41f6dc6f42f974c .

PR #29, opened 2018-05-17
fixup zsnes versions

ryantm:
* 1.51b is called "Experimental Linux Binaries" http://board.zsnes.com/phpBB3/viewtopic.php?t=11513
* 1.510 all the debian-derived repositories seem to have an extra 0 on the end of the package version string


ryantm:
@AMDmi3  Thanks for the feeback, and sorry about the alphabetizing, I meant to do that but messed it up. I think I fixed everything you wanted.

PR #32, opened 2018-05-31
fix for jade-dashboard

codesardine:


AMDmi3:
Shouldn't it be merged into `jadesktop` instead? By the name of the repo.

PR #33, opened 2018-05-31
bind: odd minor versions will be development

ryantm:


AMDmi3:
I think this needs `verge` condition to not affect past versions.

AMDmi3:
Also mistake in pattern, see Travis.

ryantm:
@AMDmi3 fixed, thanks for the feedback.

PR #42, opened 2018-07-03
add xcircuit development exception

ryantm:


ryantm:
@AMDmi3 I am hopeful this won't require maintenance because of my exchange with the developer: http://opencircuitdesign.com/pipermail/xcircuit-dev/2018-June/000986.html

PR #45, opened 2018-07-06
Merge kcm-wacomtablet and wacomtablet

Chemrat:
KDE has a project awry called "wacomtablet" (https://cgit.kde.org/wacomtablet.git/), so some distributions add kcm- prefix to the package (KDE Config Module), since it's not obvious that it belongs to KDE

### DIFF
--- a/800.renames-and-merges/kde.yaml
+++ b/800.renames-and-merges/kde.yaml
@@ -207,3 +207,6 @@
     )(?:4|5|-kde4|-kde5|-kf5|-light|-qt5)?
   setname: "$1"
 
+# KDE has a project awry called "wacomtablet", so some distributions add kcm- prefix
+# to the package (KDE Config Module), since it's not obvious that it belongs to KDE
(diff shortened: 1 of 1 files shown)


AMDmi3:
Is the name with `kcm-` prefix really preferred? Cause `wacomtablet` seem to be informative enough and also more popular. If there're actually any clashes with unrelated packages with the same name, the rule would belong to 850 part of the ruleset instead. However if prefix is somewhat official, I guess it should be preferred.

Chemrat:
The prefix isn't exactly official, I've assumed it's preferred in repology for the same reasons some distributions decided to add it. For Gentoo and Neon such name isn't a big deal since Gentoo has categories and Neon doesn't package anything but KDE. Only other distro that has an up to date package by it's original name is Mageia.

I don't think there are any clashes for now, so probably you're right and it's better to use the original name. Should I swap them?

AMDmi3:
Well let's merge it as is for now.

PR #47, opened 2018-07-10
Add kcm-tablet to kcm-wacomtablet rename rule

Chemrat:
openSuSE-specific package name: https://build.opensuse.org/package/show/openSUSE:Leap:15.0:Update/kcm_tablet


AMDmi3:
@Chemrat it turns out it should've been `kcm_tablet` (e.g. with underscore). Could you please fix it?

PR #48, opened 2018-07-11
Fix kcm-tablet -> kcm_tablet in wacomtablet rule

Chemrat:


AMDmi3:
Thanks! Not sure how the error shown by Travis has crawled in, it's unrelated.

PR #50, opened 2018-08-01
Split two different Lua LibYAML binding projects

alerque:


alerque:
Okay I'l adjust that.

AMDmi3:
Version must be a string, not an integer.

alerque:
Oops, sorry I should have run `make check on that myself. Fixed.

AMDmi3:
Thanks!

PR #53, opened 2018-08-04
Merge font packages into appropriate meta packages

alerque:
I'm not sure about the Amiri related commit in here. This is a two stage problem: first the CTAN / texlive crew have to pickup the latest upstream font, then distros have to package the package. Lumping them together is interesting from a standpoint of seeing how the font is trickling down into distros, but from a packager standpoint it might show an out of date package that is not the maintainer's problem: the CTAN upstream could be at fault.

How have you handled that case before? Should I pull that commit out of this PR?

### DIFF
--- a/800.renames-and-merges/fonts/l.yaml
+++ b/800.renames-and-merges/fonts/l.yaml
@@ -34,6 +34,12 @@
     - ttf-liberation
     - ttf-liberation-sans-narrow
 
+- setname: "fonts:libertinus"
+  name:
+    - libertinus
+    - libertinus-fonts
+    - otf-libertinus
+
 - setname: "fonts:linex"
   name:
     - fonts-linex
--- a/800.renames-and-merges/fonts/m.yaml
+++ b/800.renames-and-merges/fonts/m.yaml
@@ -77,6 +77,14 @@
     - montecarlo-font
     - montecarlo_fonts
 
+- setname: "fonts:montserrat"
+  name:
+    - fonts-otf-julietaula-montserrat
+    - julietaula-montserrat-fonts
+    - montserrat
+    - otf-montserrat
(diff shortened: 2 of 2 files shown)


800.renames-and-merges/fonts/l.yaml
```diff
@@ -34,6 +34,13 @@
     - ttf-liberation
     - ttf-liberation-sans-narrow
 
+- setname: "fonts:libertinus"
+  name:
+    - libertinus
+    - libertinus-fonts
+    - otf-libertinus
+    - otf-libertinus-git
```
AMDmi3: No need for this one, as `-git` suffix in AUR is handled globally.

800.renames-and-merges/fonts/a.yaml
```diff
@@ -80,6 +80,7 @@
     - amiri-fonts
     - fonts-hosny-amiri
     - fonts-ttf-amiri
+    - texlive-amiri
```
AMDmi3: I don't think this is correct. Texlive packages are not the same as plain fonts packages.

800.renames-and-merges/fonts/m.yaml
```diff
@@ -77,6 +77,13 @@
     - montecarlo-font
     - montecarlo_fonts
 
+- setname: "fonts:montserrat"
+  name:
+    - julietaula-montserrat-fonts
+    - montserrat
+    - otf-montserrat
+    - ttf-montserrat
```
AMDmi3: Looks like it's missing `fonts-otf-julietaula-montserrat` from Sisyphus.

alerque:
Thanks for the tips. I've dropped the commit relating to Amiri's CTAN package and edited the other ones per your notes.

PR #54, opened 2018-08-04
Combine various Gitlab packages that are release-synced

alerque:
Can you review the CE vs. EE comment in the commit here? Is this the way to handle a split release like this where the same base code is released with two licenses and some package differences upstream but that are always version-synced?


AMDmi3:
This looks like it needs some [flavors](https://github.com/repology/repology-rules#addflavor), so if `gitlab-ee`/`gitlab-ce` happen to have different versions, older one would be counted as _outdated_ and not _legacy_. You can just add `addflavor: true` to both rules.

alerque:
Like that?

AMDmi3:
Yes, but you don't need to duplicate target name in the `name:` list.

alerque:
Dropped.

PR #59, opened 2018-08-07
Fix dated snapshot anomaly

alerque:
This should fix the irregularly patterned date format missed in #52.


AMDmi3:
Let's leave first rule with `.*`, wider regexp won't hurt in this case.

alerque:
> Let's leave first rule with `.*`, wider regexp won't hurt in this  
> case.

Done.

AMDmi3:
I've just noticed that this was not really needed. This version is already considered outdated, so there's no need to ignore it as it doesn't affect status of correct versions. Ignoring it is a no-op as well.

PR #60, opened 2018-08-07
Ignore bad lyricue version in Ubuntu universe

alerque:
Somebody did an oops and bumped the version number instead of the package build number. 4.0.12 is the lest release, except for on Ubuntu where an otherwise unknown release 4.0.13 fixes a packaging issue, but is actually 4.0.12 code underneath.

I don't know how to fix this properly (compare 4.0.13 as if it was 4.0.12 if labeled "is really"). Is there a way to parse this string and use the "actual" number instead of the reported one for comparison purposes? E.g. parse `4.0.13isactually4.0.12` as if it was `4.0.12`?


AMDmi3:
You could use `ver: "4.0.13isactually4.0.12", server: "4.0.12"` or `verpat: ".*isactually(.*)", setver: $1`, but such a mess needs a clear indication of that it's incorrect, at least until there's a way to create problems with rules.

PR #61, opened 2018-08-07
Merge Mattermost packages by flavor

alerque:
Note the `mattermost-webapp` & `mattermost-server` packages are synchronized upstream and distributed in a single combined `mattermost` release, some downstream packagers split them out again, hence the flavors. On the other hand `mattermost-desktop` is on a separate release schedule.


AMDmi3:
Please don't mix unrelated changes in a single PR. Mattermost rule is OK apart from syntax error. Gentoo rule should be checked for that it does not result in mismerges first.

PR #62, opened 2018-08-07
Catch all Gentoo *-bin ebuilds as flavors

alerque:
Split out of #61 — 

> Also note that the AUR `*-bin` rule caught [mattermost-desktop-bin](https://repology.org/metapackage/mattermost-desktop/versions) for AUR, but not for [Gentoo or Funtoo](https://repology.org/metapackage/mattermost-desktop-bin/versions). I added a similar rule for Gentoo, but I wasn't sure what to use for Funtoo because there are zero examples of rules involving it in the set. Does it inherit from Gentoo?


AMDmi3:
Funtoo uses `gentoo` ruleset so it'll affect both. I'll check that it doesn't cause any problems and merge it.

AMDmi3:
Merged in e09204499f5aabb2fe8f813f6833dfce42a5c911

PR #66, opened 2018-08-10
Merge Brave browser packages

alerque:
In spite of the "snapshot" in the name of one of these, all of these are
packaging release tags. None of them are actually snapshots or even
different flavors, just different packaging.


800.renames-and-merges/b.yaml
```diff
@@ -102,6 +102,7 @@
 - { setname: box2d,                    name: libbox2d }
 - { setname: bpython,                  name: [bpython-py27,bpython2] }
 - { setname: brasero,                  name: [brasero2,brasero-dvdcssfix] }
+- { setname: brave,                    name: [brave-bin, brave-browser, brave-snapshot] }
```
AMDmi3: `brave-bin` needs `addflavor`, and `brave-snapshot` needs `weak_devel: true`, so there should be 3 rules.

`weak_devel` still needs to be documented, but see repology/repology#654.

alerque: Done.

PR #67, opened 2018-08-10
Fix Apache versions schemes

alerque:
Packages labeled apache2 are likely supposed to be 2.4, but packages
with names of apache20 or apache22 are likely legacy, not outdated.

Meanwhile apache25 is devel. There is a catch all rule for odd numbered
versions, but it isn't catching e.g. the AUR apache25 package.

### DIFF
--- a/900.version-fixes/a.yaml
+++ b/900.version-fixes/a.yaml
@@ -35,6 +35,9 @@
 - { name: aspcud,                      ver: "2011.03.17",                                  ignore: true }
 - { name: assaultcube,                 verpat: "[0-9]{8}",                                 ignore: true }
 - { name: apache,                      verpat: "2\\.[0-9]*[13579]\\..*",                   devel: true }
+- { name: apache20,                                                                        legacy: true }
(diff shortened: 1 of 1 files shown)


alerque:
Is matching names here that have already had `setname` run on them to normalize them even supported? I don't have a way of checking what this does to the data, I'm kind of shooting blind here and this was a guess.

AMDmi3:
This is not correct in multiple ways. First, you can't match names here, as all `apache*` became `apache` in 800 ruleset.  The general logic is merge (800) -> split (850) -> version rules (900) -> splits based on version (950). Next, you don't need to mark older versions legacy in generic case, as it's done automatically, as long as newer version exists. That is, if 2.4 exists in the repo, 2.2 in the same repo would be legacy. Otherwise, it'd be outdated, which is true as the latest version is not there. Finally, the devel rule does catch AUR package. It's just compared older than 2.5.0-alpha

AMDmi3:
Time out

PR #70, opened 2018-08-30
Teamocil is a shell utility that happens to be implemented in Ruby

alerque:


AMDmi3:
Apart from an invalid regexp, that's not how we do it. Not using global rules here is a bad idea as you'd still need to merge all ruby prefixes (and many repositories still use them even for applications, such as FreeBSD's `rubygem-teamocil`) manually. Besider, the regexp is too wide and may catch unrelated projects in the future. So instead, please add a merge rule which merges `ruby:teamocil` into `teamocil`.

alerque:
Like that?

AMDmi3:
Exactly. Thanks!

PR #73, opened 2018-09-05
Fix two issues with dmraid versioning

awilfox:
* Mageia, OpenMandriva, PCLinuxOS, and PLD put 'rc16' in the 'pkgrel'
  field instead of the version, so it looks like they are shipping 1.0.0.
  Such a version does not exist.

* Arch and its derivatives, KaOS, and SliTaz add an extraneous '.3' or
  '-3' to the version number.  This appears to be an Arch patchset and
  has nothing to do with upstream.

(Hopefully I have understood verpat correctly.  If not, please comment and I will amend this.)

### DIFF
--- a/900.version-fixes/d.yaml
+++ b/900.version-fixes/d.yaml
@@ -38,6 +38,7 @@
 - { name: distorm,                     verpat: "20[0-9]{6}.*",                             ignore: true } # freebsd garbage
 - { name: djvulibre,                   ver: "3.5.27.1",                                    ignore: true }
 - { name: dkms,                        ver: "2.022.0",                                     ignore: true } # gobolinux fake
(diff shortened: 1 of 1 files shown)


AMDmi3:
Verpat is used correctly, although I'd use `1\\.0\\.0[._-]?rc16[._-]3`, first bracket group to catch more possible variants and be more future proof and second one mainly for explicitness.

However rc16-3 thingy does seem to be official. See people.redhat.com/~heinzm/sw/dmraid/src and people.redhat.com/~heinzm/sw/dmraid/src/old.

awilfox:
Good catch.  Ok, it only does the Mageia/etc fix-up now.

PR #92, opened 2018-10-25
Merge three packages from OpenSUSE repos into their appropriate metapackages

s0600204:
The various OpenSUSE repos each have packages as follows:
* `mozilla-jss`
* `mozilla-nspr`
* `mozillafirefox`
* `mozillafirefox-branding-opensuse`

The first three appear to be equivalent to the metapackages `jss`, `nspr`, and
`firefox` respectively.

This PR amends one and adds two rename rules so as to merge the first three into their correct metapackages, whilst leaving the fourth intact.

----

(An alternate approach would be a single OpenSUSE-specific rule that removes the `mozilla` prefix from all four. As there already exists a metapackage called `firefox-branding-iceweasel`, one called `firefox-branding-opensuse` wouldn't seem too out of place.)


AMDmi3:
Please fix syntax, there are missing commas.

s0600204:
My apologies.

AMDmi3:
:+1: thanks!

PR #94, opened 2018-10-27
add version fix for rstudio 1.2 previews

ciil:
The RStudio 1.2 releases are currently part of a "preview" line (https://www.rstudio.com/products/rstudio/download/preview/ vs. the official 1.1 release line: https://www.rstudio.com/products/rstudio/download/). I hope tagging it as devel is the right move here. If not, just tell me.


AMDmi3:
Please add `maintenance: true`, as it seems that some later 1.2.xxx version is going to become stable and the rule would need to be updated. Also `devel: true` looks misaligned by 4 spaces, please fix.

PR #143, opened 2019-01-12
Merge environment-modules

xdelaruelle:
Merge the different names the [Modules](http://modules.sf.net) project has out there. `environment-modules` is preferred as main name for disambiguation.

### DIFF
--- a/800.renames-and-merges/e.yaml
+++ b/800.renames-and-merges/e.yaml
@@ -67,6 +67,7 @@
 - { setname: enlightenment,            name: e17 }
 - { setname: enscript,                 namepat: "enscript-.*", ruleset: freebsd } # enscript-a4, enscript-letter, enscript-letterdj
 - { setname: entagged,                 name: entagged-tageditor, ruleset: gentoo }
+- { setname: environment-modules,      name: [modules,env-modules,env-modules-tcl] }
 - { setname: eom,                      name: eom-dev, weak_devel: true }
 - { setname: eom,                      name: eom-gtk2, addflavor: true }
 - { setname: eot-utils,                name: eot-utilities }


800.renames-and-merges/e.yaml
```diff
@@ -67,6 +67,9 @@
 - { setname: enlightenment,            name: e17 }
 - { setname: enscript,                 namepat: "enscript-.*", ruleset: freebsd } # enscript-a4, enscript-letter, enscript-letterdj
 - { setname: entagged,                 name: entagged-tageditor, ruleset: gentoo }
+- { setname: environment-modules,      name: modules }
```
AMDmi3: This could be merged with the next line

800.renames-and-merges/e.yaml
```diff
@@ -67,6 +67,9 @@
...
+- { setname: environment-modules,      name: env-modules }
+- { setname: environment-modules,      name: env-modules-tcl }
```
AMDmi3: Please add `addflavor: tcl` flag

800.renames-and-merges/e.yaml
xdelaruelle: done

800.renames-and-merges/e.yaml
xdelaruelle: Understood. `env-modules-tcl` corresponds to the regular version 4 of Modules and `env-modules` to a compatibility version. So I have added a `compat` flavor to the `env-modules` name.

AMDmi3: Hmm, in this case flavor is not needed, I was misled by the package naming. Flavor is only needed to distinguish subpackages or variant packages, but not packages which only differ by version.

xdelaruelle: Ok, so 1 rule to rename all 3 names to 'environment-modules'

PR #184, opened 2019-05-01
Mark AOSC recommonmark release as a Git snapshot

awilfox:


AMDmi3:
The common way to deal with these is to use e.g. `verpat: "20[0-9]{6}"`, as it's likely to be updated with another date rather sooner than later.

AMDmi3:
Please fix the issue found by Travis.

PR #187, opened 2019-05-01
Add renames for packages that Adélie names differently

awilfox:


800.renames-and-merges/python.yaml
```diff
@@ -115,6 +115,7 @@
 - { setname: "python:khal",            name: khal }
 - { setname: "python:kiwisolver",      name: kiwisolver }
 - { setname: "python:lektor",          name: lektor }
+- { setname: "python:lesscpy",         name: lesscpy }
```
AMDmi3: It looks like lesscpy would be a more proper target name, as this is primarily an application and not a python module.

PR #199, opened 2019-05-22
Delete rstudio preview version-fixes rule

ciil:
The rule introduced in https://github.com/repology/repology-rules/pull/94 can now be discontinued with the release of rstudio 1.2 as the mainline/stable release (https://www.rstudio.com/products/rstudio/release-notes/).


AMDmi3:
Isn't there going to be next 1.3 development branch?

ciil:
I haven't heard anything about that, I think it might've been a one-off thing, but I'm just going to ask over there, so don't merge yet.

AMDmi3:
To stay on the future-proof side, I'll just bump version in devel rule. Some insight from the upstream would still be useful.

ciil:
I've asked [upstream](https://community.rstudio.com/t/versioning-for-rstudio-preview-releases/31675), and the current policy seems to be to usually just have preview releases with an increased patch/build version. 1.3 preview releases aren't expected until the end of the year, but will definitely happen, so this is probably the best we can do for the time being.

PR #201, opened 2019-05-27
Initial rules additions for Cygwin

jon-turney:


850.split-ambiguities/p.yaml
```diff
@@ -5,6 +5,7 @@
 
 - { name: pal, wwwpart: palcal, setname: palcal }
 - { name: pal, wwwpart: starlink, setname: pal-astronomical }
+- { name: pal, category: utils, setname: palcal } # cygwin
```
AMDmi3: Better use `ruleset:` here, as utils is too generic and repository is implied anyway.

800.renames-and-merges/t.yaml
```diff
@@ -92,6 +92,7 @@
 - { setname: tint,                     name: tint-tetris }
 - { setname: tintin++,                 name: [tintin,tintinxx] }
 - { setname: tintin++,                 name: tintin-debug, addflavor: true }
+- { setname: tix,                      name: tcl-tix }
```
AMDmi3: `tk-tix` could be added too

800.renames-and-merges/s.yaml
```diff
@@ -29,6 +29,7 @@
 - { setname: schilytools,              name: schily }
 - { setname: schismtracker,            name: schism }
 - { setname: schroedinger,             name: libschroedinger }
+- { setname: schroedinger,             namepat: "schroedinger[0-9.-]*" }
```
AMDmi3: Better use `+` instead of `*` here, so there's no no-op match.

800.renames-and-merges/ruby.yaml
```diff
@@ -117,6 +117,7 @@
 - { setname: "ruby:bundler",           name: bundler }
 - { setname: "ruby:bundler",           namepat: "ruby[0-9.]+-bundler" }
 - { setname: "ruby:gems",              name: ["ruby:rubygems",ruby-gems,rubygems] }
+- { setname: "ruby:goocanvas",         namepat: "ruby-goocanvas[0-9.]+" }
```
AMDmi3: Since the package was matched by wildcard rule for ruby packages before, it'll be named `ruby:goocanvasXXX` when this rule is checked.

jon-turney:
> * Let's split this into two PRs: the global part (wildcard rules, checker update) and the part for individual package rules. It would be also more convenient to split the latter to commit-per-package, so they can be cherry-picked
> 
> * For rules which merge package names with version appended don't hesitate to use wide and uniform regexp pattern, e.g. `libunique[0-9.-]+`, `libconfuse[0-9.-]+`, `gobby[0-9.-]+`, `glib[0-9.-]+-networking`, `clutter[0-9.-]+` etc.

Updated and split off as #202

jon-turney:
>     * Wildcard rule for python modules seem to be missing. Likely the `python3?-(.*)` one.

I've added the cygwin ruleset to the list in the 'python-(.*)' rule, which I think is the correct one (after a fix to the parser)

jon-turney:
Thanks. It didn't occur to me to look at the 'newest only in cygwin' search results for anomalies.

Re: librsb etc., the full version is '1.2.0-0.rc7a', which should probably be treated in the same way as Fedora packages with a release starting with a 0.

I attempted that in https://github.com/repology/repology/pull/859

Issue #10, opened 2018-03-20
MinGW packages


AMDmi3:
From https://github.com/repology/repology/issues/248 by @mojca:

The [MinGW packages](https://repology.org/metapackages/all/?search=mingw) are currently a bit of a mess, with each distribution providing its own package names.

There are two "different" projects, MinGW and MinGW-w64. And package are different:
- the compiler (gcc) which should be at some fixed version of gcc (compared to gcc versions)
- binutils (which should be compared to binutils versions)
- mingw-w64 itself
- potentially other libraries for use when cross-compiling

In case of MacPorts we have five packages that derive from mingw-w64 for example:
* mingw-w64 @5.0.2 (meta port)
* i686-w64-mingw32-crt @5.0.2
* i686-w64-mingw32-headers @5.0.2
* x86_64-w64-mingw32-crt @5.0.2
* x86_64-w64-mingw32-headers @5.0.2

plus binutils, gcc, ... as well as the old MinGW package.

AMDmi3:
> The MinGW packages are currently a bit of a mess

True.

> There are two "different" projects, MinGW and MinGW-w64

I don't have a clear understanding of difference between these yet and thus I don't know whether these could/should be merged or not.

Either way, we need to:
- merge toolchain parts (gcc, binutils, gdb) into either
  - their originals (gcc, binutils, gdb)
  - a distinct set of mingw packages
  - a two distinct sets of packages for mingw and mingw-w64
- one or to metapackages for the library (crt/runtime/headers) correspondingly
- merge all libraries with their original metapackages (mingw-w64-boost -> boost), but we need flavors (repology/repology#170) for that first

jon-turney:
This is a specific case of a slightly more general problem, packages which contain tools or libraries for cross-compiling.

> > There are two "different" projects, MinGW and MinGW-w64
> 
> I don't have a clear understanding of difference between these yet and thus I don't know whether these could/should be merged or not.

I don't think MinGW and MinGW-w64 runtime/header version numbers are meaningfully comparable, so I think they should be separate.

AMDmi3:
> This is a specific case of a slightly more general problem, packages which contain tools or libraries for cross-compiling.

Yes.

> I don't think MinGW and MinGW-w64 runtime/header version numbers are meaningfully comparable, so I think they should be separate.

I currently have no idea how these are versioned. This issue is about merging cross-development libraries for mingw, which are (usually) versioned the same way as their native counterparts:

[...]

PR #210, opened 2019-06-18
Normalize Cygwin cross-package names

jon-turney:
This normalizes the names of packages for x86/x86_64 Cygwin cross-compilation.

This probably isn't the right way to do this, and a proper solution depends on https://github.com/repology/repology/issues/170 and https://github.com/repology/repology-rules/issues/10.

Note that there are also some djgpp cross-packages, which also need similar handling.


AMDmi3:
> This probably isn't the right way to do this

No, this is fine, as I plan do to the same with mingw cross-packages as well. There's no way to store information on cross-development yet, but we can introduce it anytime later.

PR #221, opened 2019-07-16
Set exherbo/void browserpass as browserpass-native

Johnnynator:
The exherbo package is the outdated native part of browserpass-legacy
The void package is the same as other browserpass-native packages,


AMDmi3:
Btw, you could fix this by updating URL in void package, for https://github.com/browserpass/browserpass redirects to https://github.com/browserpass/browserpass-legacy

Johnnynator:
> Btw, you could fix this by updating URL in void package, for https://github.com/browserpass/browserpass redirects to https://github.com/browserpass/browserpass-legacy

thx for pointing this out, should point to the correct repo now.

PR #222, opened 2019-07-16
Add verignore for multibootusb

Johnnynator:
I don't have any clue where this version comes from both Sourceforge and Github only have 9.2.0 as newest version.


AMDmi3:
You can flag entries as incorrect on http://freshcode.club, they are usually fixed quite fast. I've done so for this one.

AMDmi3:
Okay, they no longer react to flagging projects.

PR #232, opened 2019-08-05
Sort out OpenFOAM

### LINKED ISSUE #231: OpenFoam
There is a project I use, OpenFOAM.

The problem is, [versioning is a mess](https://repology.org/projects/?search=openfoam&maintainer=&category=&inrepo=&notinrepo=&repos=&families=&repos_newest=&families_newest=). 

OpenFOAM has two separate forks:

  - Fork maintained at [openfoam.org](https://openfoam.org/), had versioning scheme `x.y.z`, then `x.y`, and now just `x`

  - Fork maintained at [openfoam.com](https://openfoam.com/) by company ESI, has verison scheme `vyymm` such as `v1906`. This fork in called `openfoam-esi` and `openfoam-plus` in some distributions, but has generic name `openfoam` in others mixing it with the first one. 

What would be a correct approach on writing rules for OpenFOAM? We need the following:

  - Choose one name for the second fork. I think `openfoam-plus` is better because this is how the repository is called (https://develop.openfoam.com/Development/OpenFOAM-plus) and also it would not become obsolete if the fork changes hands. (ESI is the company currently owning it)
  - Rename packages with `yymm` version to `vyymm` version.
  - Rename `openfoam` packages with `vyymm` version scheme to `openfoam-plus`

In which files should I write these rules? Is there a way to test the rules easily?

ilyapopov:
- Consistently use name `openfoam-plus` for openfoam.com fork
  - Rename packages with `yymm` version scheme to `openfoam-plus`
  - yymm -> vyymm

Fixes #231

### DIFF
--- a/800.renames-and-merges/o.yaml
+++ b/800.renames-and-merges/o.yaml
@@ -55,6 +55,7 @@
 - { setname: opendnssec,               name: opendnssec2 }
 - { setname: openerp,                  namepat: "openerp[0-9.-]+" }
 - { setname: openfoam,                 namepat: "openfoam[0-9x.-]+" }
+- { setname: openfoam-plus,            namepat: "openfoam-esi.*" }
 - { setname: openhab,                  name: openhab-beta, weak_devel: true }
 - { setname: openhab,                  name: openhab2 }
 - { setname: openhab-addons,           name: openhab2-addons }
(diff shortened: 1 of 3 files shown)


AMDmi3:
`vyymm` is incorrect, `v` is never part of the version. Please remove it.

AMDmi3:
Timeout

PR #238, opened 2019-08-11
shotcut: set v19.08.05 as devel

Ikke:
[v19.08.05][0] is marked as BETA / pre-release

[0]:https://github.com/mltframework/shotcut/releases/tag/v19.08.05


AMDmi3:
`v` is not part of the version. Also please add `maintenance: true` to denote that this rule should be manually updated.

Ikke:
I've removed the 'v' prefix and set the maintenance attribute.

PR #239, opened 2019-08-11
alpine: set v2.21.9999 as devel

Ikke:
[v2.21.9999][0] is an alpha release.

[0]:http://alpine.x10host.com/alpine/alpha/release/VERSION.html


AMDmi3:
Please fix syntax. And I'd use `verpat: ".*99"` instead to make it future proof.

Ikke:
@AMDmi3 Thanks. Fixed the syntax error and changed the rule to use the version pattern.

PR #240, opened 2019-08-12
Clean up xfce4 dupes

luzpaz:


AMDmi3:
I'd rather have a generic rule which normalizes all of these at once. Probably something similar to what we have in `800.renames-and-merges/kde.yaml`.

luzpaz:
@AMDmi3 Agreed, please lead the way and I'll study how you create this. The `kde` example is complex and I'm having trouble following it.

AMDmi3:
Yes. But more like (btw, there's no `netwatch`):

```
- namepat: |
    xfce4-(?:panel-plugin-)?(
    mailwatch|
    mount|
    mpc
    )(?:-plugin)?
  setname: xfce4-$1-plugin
```

as it's done with kde. The idea is to catch all patterns (as I see, there are currently `xfce4-XXX`, `xfce4-XXX-plugin` and `xfce4-panel-plugin-XXX`), but specify items by hand, as otherwise we'll have to use `xfce4-(.*)` pattern which would be too wide. And the formatting improves readability (especially for diffs) and allows sorting (but watch for last entry without `|`).

PR #241, opened 2019-08-13
Merge for FreeBSD elm-format

avh4:
This is intended to merge https://repology.org/project/haskell:elm-format/versions into https://repology.org/project/elm-format/versions


AMDmi3:
Good, but please add it to existing `800.renames-and-merges/e.yaml`. Separate files are used only under special circumstances.

avh4:
Updated!   I also created https://github.com/repology/repology-rules/pull/249 to try to clarify the README a bit

PR #242, opened 2019-08-13
Add Adélie php rule and fix "Apache Lucene" from Wikidata

sroracle:
* Add a rule to rewrite [e.g. `php7-apcu`](https://repology.org/project/php7-apcu/versions) and any future packages to `php:` namespace for Adélie.
* Rewrite "Apache Lucene" to "lucene" so that [Wikidata doesn't corrupt the current version of clucene](https://repology.org/project/clucene/versions).


AMDmi3:
Please remove lucene related commit, it will be fixed in wikidata (https://phabricator.wikimedia.org/T228640#5411245).

AMDmi3:
On the second glance, there's only one package which falls under this rule. Is it safe to assume that it'll match more packages in the future?

sroracle:
Yes. Right now there's only apcu, but any future PEAR/PECL packages will also fall under `php7-` or `php8-` or `php-`.

Rebased onto master.

PR #245, opened 2019-08-18
split gip package

jcgruenhage:


AMDmi3:
Split rules based on `ruleset` condition are discouraged and used as a last resort. You could use e.g. `wwwpat: dalance` here instead.

PR #250, opened 2019-08-21
merge comrak and rust:comrak

jcgruenhage:


AMDmi3:
Is this primarily a module or an application? Either way, you don't need to merge rust-comrak, it's already merged.

jcgruenhage:
comrak (as packaged for void) is an application, but you can also use it as a library. Since rust is statically linked usually, most of the rust based packages in repos these days are going to be applications, even when those things can also be used as a module, even when the primary use case is being used as a module.

AMDmi3:
Ok, then please just remove `rust-comrak` from names.

PR #256, opened 2019-08-26
Set flavors for older versions of elm-format

avh4:
Here's what the elm-format versions currently look like:

Screen Shot 2019-08-26 at 4 36 25 PM

This PR is meant to take for instance the `elm-format-0.18-bin` package and give it the flavor "elm0.18"  (it indicates which version of Elm-platform that particular build of elm-format supports).
(as of elm-format 0.8.1, there are no longer flavors; the same elm-format binary works for all versions of Elm)

### DIFF
--- a/800.renames-and-merges/e.yaml
+++ b/800.renames-and-merges/e.yaml
@@ -40,7 +40,8 @@
 - { setname: elixir,                   name: elixir-lang }
 - { setname: elkcode,                  name: elk-chemistry }
 - { setname: elm-format,               name: "haskell:elm-format" }
-- { setname: elm-format,               namepat: "elm-format[0-9.-]+(?:-exp)?" }
+- { setname: elm-format,               namepat: "elm-format(?:-exp)?(?:-bin)?" }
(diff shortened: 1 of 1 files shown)


AMDmi3:
What are you trying to achieve with this?

AMDmi3:
> I was assuming that would have some effect on how all the orange rows are shown?

Yes, this would effectively make them outdated, is this expected? I'm confused because AUR people usually desire the opposite, as it's common to have old versions preserved in AUR with no recent version available, and they don't want these to be reported as outdated. We have facilities for this as well (220 ruleset).

If it's actually desired, it'd be better to just add `addflavor: true` to the original rule - it takes care of unique flavors for you, and you don't need to handle `-bin` as it's already handled globally in 100 ruleset. 

>  I guess I was hoping to make the list of versions there be less cluttered

This will only affect statuses. There's related repology/repology#711, but I don't think it will affect this case either, as it's not supposed to conflate rows which differ in multiple columns, and Repology is not supposed to hide any packages. IMO there's no problem in that all of them are shown.

PR #272, opened 2019-09-07
Merge foxdot

luzpaz:


AMDmi3:
It doesn't look like a modules, so it should be the other way around

luzpaz:
reverted, amended, and repushed

PR #281, opened 2019-09-21
Merge spyder

luzpaz:


AMDmi3:
Could you please quote `python:spyder`? Seems to be OK in this case, but generally colon means a mapping in YAML which would be incorrect syntax here.

luzpaz:
Done

PR #282, opened 2019-09-26
Set proper last version for salt

Ikke:
Salt official releases have a 3-part version scheme, but they also push
2-part tags to github, which are not official releases.

For example the last official release is 2019.2.1. 2019.8 is
available as tag on github, but not as release (and also not available
on pypi, which they consider the official source for releases[0]).

[0]:https://github.com/saltstack/salt/issues/41847


AMDmi3:
`last` should not used in the normal ruleset. It looks like you don't need a rule for official versions at all, just a `devel` one is fine.

Ikke:
@AMDmi3 Thanks, I've removed the `last` rule.

PR #305, opened 2019-11-28
Merge claws-mail plugins

niacat:
These plugins are distributed as part of the claws-mail distribution so the version number is the same as the main package. Some packaging systems have decided to split them out from claws-mail.


AMDmi3:
These kinds of splits are common in pkgsrc, it's a burden to merge them all manually. I'd prefer to have a way to know master port for such splits.

PR #306, opened 2019-11-28
openttd-data is a duplicate of openttd-opengfx

niacat:


AMDmi3:
This is not correct, the same way as `openttd-data` is not correct - it packages multiple projects with their own versions, hiding them behind (randomly chosen? greatest? openttd?) version and providing no way to know versions of individual components.

PR #321, opened 2020-01-12
Update o.yaml for openkim-models

ellio167:
openkim-models: convert YYYYMMDD to YYYY.MM.DD for consistency.

### DIFF
--- a/900.version-fixes/o.yaml
+++ b/900.version-fixes/o.yaml
@@ -64,7 +64,7 @@
 - { name: openjazz,                                                  ruleset: openbsd,     untrusted: true } # accused of fake 20190120
 - { name: openjfx,                     verpat: "8\\.0\\..*",                               ignore: true } # fedora; the official scheme is e.g. 8u151b12
 - { name: openjk,                                                                          noscheme: true }
-- { name: openkim-models,              verpat: "20[0-9]{6}",                               incorrect: true } # it's YYYY-MM-DD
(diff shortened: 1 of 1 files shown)


AMDmi3:
We don't fix others errors (incorrect version in this case) in Repology. This should be fixed in Debian.

ellio167:
@AMDmi3 I'm confused, there seem to be many examples of this sort of thing in the repo.

`c.yaml:- { name: cmst,                        verpat: "(20[0-9]{2})([0-9]{2})([0-9]{2})",         setver: "$1.$2.$3" } # manjaro`

`t.yaml:- { name: tuxpaint-stamps,             verpat: "(20[0-9]{2})([0-9]{2})([0-9]{2})",         setver: "$1.$2.$3" }`

It seems that the only way to fix this issue in Debian is to increment the `epoch` so that the new version of the package would be `1:2019.03.31`.  Will repology handle this cleanly and identify `1:2019.03.31` as equivalent to `2019.03.31`?

AMDmi3:
> I'm confused, there seem to be many examples of this sort of thing in the repo.

Some are legacy, some are inevitable because of unclear upstream versioning scheme. But the default policy is highlighting problems in repositories instead of masking them.

> Will repology handle this cleanly and identify 1:2019.03.31 as equivalent to 2019.03.31

Sure, dropping epochs and revisions is the first thing repology does when parsing a version.

ellio167:
OK, great.  We'll bump the epoch.  Thanks....

PR #332, opened 2020-01-28
OS4Depot: disambiguate tilemaker

simon04:


AMDmi3:
There's a dedicated blacklist for amiga-only software with clashing names, let's better put it there. See `200.blacklist/os4depot.yaml`.

simon04:
Done via  6b560a3.

PR #344, opened 2020-03-09
Ignore versions for rtl8821ce

Ikke:
Upstream has no defined versions.


AMDmi3:
Shouldn't it be 8821?

Ikke:
Yes, thank you.

PR #354, opened 2020-04-03
New rules for Carbs Linux

cemkeylan:
These patches,

- Specify a seperate package manager wwwpart for kiss-package-manager
- Imitate the versioning rule for git packages


AMDmi3:
Looks good, but please squish it into a single commit

PR #381, opened 2020-06-18
Fixing wrong versions

TimB87:
Hi @AMDmi3 

I would like to fix some generic rules, like these ones. This PR will be marked as a draft to fill it up and have a place to ask about specific rules, if that's ok with you?
In a second step I would like to go in and check for CRUX specifc rules, but that would be seperate pr as I see it.

Cheers

### DIFF
--- a/900.version-fixes/f.yaml
+++ b/900.version-fixes/f.yaml
@@ -1,5 +1,7 @@
 # vim: tabstop=39 expandtab softtabstop=39
 
+- { name: f.lux,                       verpat: "[4-9]{1}\\..*",                             incorrect: true } # latest is version 1.2.0.. where does 4.x come from?
+- { name: f.lux,                       ruleset: chocolatey,                                untrusted: true } # sorry chocolatey
 - { name: f3,                          verpat: ".*unstable",                               ignore: true } # nix garbage
 - { name: facetimehd,                                                                      noscheme: true }
 - { name: fahclient,                   verlonger: 3,                                       incorrect: true }
@@ -88,8 +90,6 @@
 - { name: fluid-soundfont,                                           ruleset: [slackbuilds,macports], untrusted: true }
 - { name: fluidsynth-dssi,             ver: "1.9.9",                 ruleset: opensuse,    incorrect: true }
 - { name: fluidsynth-dssi,                                           ruleset: opensuse,    untrusted: true } # accused of fake 1.9.9
-- { name: f.lux,                       verpat: "[4-9]{1}\\.*"                              incorrect: true } # latest is version 1.2.0.. where does 4.x come from?
-- { name: f.lux,                       ruleset: chocolatey,                                untrusted: true } # sorry chocolatey
 - { name: flvstreamer,                 ver: "2.1.c.1",                                     setver: "2.1c1" }
 - { name: flvstreamer,                 ver: "2.1c",                                        setver: "2.1c0" }
 - { name: flvstreamer,                 ver: "2.1",                   ruleset: pisi,        incorrect: true } # 2.1c1
--- a/900.version-fixes/f.yaml
+++ b/900.version-fixes/f.yaml
@@ -88,6 +88,8 @@
 - { name: fluid-soundfont,                                           ruleset: [slackbuilds,macports], untrusted: true }
 - { name: fluidsynth-dssi,             ver: "1.9.9",                 ruleset: opensuse,    incorrect: true }
 - { name: fluidsynth-dssi,                                           ruleset: opensuse,    untrusted: true } # accused of fake 1.9.9
+- { name: f.lux,                       verpat: "[4-9]{1}\\.*"                              incorrect: true } # latest is version 1.2.0.. where does 4.x come from?
+- { name: f.lux,                       ruleset: chocolatey,                                untrusted: true } # sorry chocolatey
 - { name: flvstreamer,                 ver: "2.1.c.1",                                     setver: "2.1c1" }
 - { name: flvstreamer,                 ver: "2.1c",                                        setver: "2.1c0" }
 - { name: flvstreamer,                 ver: "2.1",                   ruleset: pisi,        incorrect: true } # 2.1c1
--- a/900.version-fixes/f.yaml
+++ b/900.version-fixes/f.yaml
@@ -1,7 +1,6 @@
 # vim: tabstop=39 expandtab softtabstop=39
 
-- { name: f.lux,                       verpat: "[4-9]{1}\\..*",                             incorrect: true } # latest is version 1.2.0.. where does 4.x come from?
-- { name: f.lux,                       ruleset: chocolatey,                                untrusted: true } # sorry chocolatey
+- { name: f.lux,                       verpat: "20[0-9]{2}\\-[0-9]{2}\\-[0-9]{2}",         incorrect: true }
 - { name: f3,                          verpat: ".*unstable",                               ignore: true } # nix garbage
 - { name: facetimehd,                                                                      noscheme: true }
 - { name: fahclient,                   verlonger: 3,                                       incorrect: true }
--- a/900.version-fixes/m.yaml
+++ b/900.version-fixes/m.yaml
@@ -80,6 +80,7 @@
 - { name: metacity,                    verpat: "[0-9]+\\.[0-9]*[13579]\\..*",              devel: true }
 - { name: metamail,                    verpat: "2\\.7\\..*",         ruleset: [gentoo,opensuse,pld], incorrect: true }
 - { name: metasploit,                                                ruleset: homebrew_casks, untrusted: true } # accused of fake 5.0.66
+- { name: metis,                                                     p_is_patch: true } # 5.1.0 < 5.1.0-p8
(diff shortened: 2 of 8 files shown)


900.version-fixes/f.yaml
```diff
@@ -1,5 +1,7 @@
 # vim: tabstop=39 expandtab softtabstop=39
 
+- { name: f.lux,                       verpat: "[4-9]{1}\\..*",                             incorrect: true } # latest is version 1.2.0.. where does 4.x come from?
+- { name: f.lux,                       ruleset: chocolatey,                                untrusted: true } # sorry chocolatey
```
AMDmi3, line 4: You don't need this, chocolatey is untrusted by default

TimB87: Indeed, actually, it's CRUX 3.4 with the wrong version scheme currently blocking the correct version detection. I will change that in a bit!

900.version-fixes/p.yaml
```diff
@@ -199,6 +199,8 @@
 - { name: prboom-plus,                                               ruleset: slackbuilds, untrusted: true } # accused of fkae 2.5.1.7um
 - { name: prelink,                     ver: "20151030",              ruleset: gentoo,      incorrect: true }
 - { name: prelink,                                                   ruleset: gentoo,      untrusted: true }
+- { name: primus,                      verpat: "20[0-9]{6}", incorrect: true } # several repos use a date, but this is a tagged release
```
AMDmi3, line 202: Having snapshots when the latest official release was half a decade ago should not be penalized by `incorrect` status.

Single `verpat: "20[0-9]{2}.*", snapshot: true` rule would be enough to ignore all dated versions, but it won't hurt to also mark `20150328` specifically as `incorrect` (since it's `0.2` judging by the date), and `20131226` as `outdated`.

900.version-fixes/perl.yaml
```diff
@@ -33,6 +33,7 @@
 - { name: "perl:date-simple",          verpat: "3\\.30\\..*",                              incorrect: true } # 3.03, not 3.30
 - { name: "perl:date-simple",          verpat: "3\\.03[._][0-9]+",                         outdated: true } # 3.03_0[123] < 3.03
 - { name: "perl:datetime",             verpat: "1\\.[0-9]{3,}.*",                          ignore: true }
+- { name: "perl:datetime-format-builder", verpat: "0\\.[0-9]{4}",                          incorrect: true }
```
AMDmi3, line 36: Let's change the regexp to `0\\.[0-9]{3}.*` to catch `0.820.0` as well

900.version-fixes/m.yaml
```diff
@@ -80,6 +80,7 @@
 - { name: metacity,                    verpat: "[0-9]+\\.[0-9]*[13579]\\..*",              devel: true }
 - { name: metamail,                    verpat: "2\\.7\\..*",         ruleset: [gentoo,opensuse,pld], incorrect: true }
 - { name: metasploit,                                                ruleset: homebrew_casks, untrusted: true } # accused of fake 5.0.66
+- { name: metis,                                                     p_is_patch: true } # 5.1.0 < 5.1.0-p8
```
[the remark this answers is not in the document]
TimB87: `- { name: metis, verpat: ".*p.*", devel: true }` doesn't help in my local instance.. and neither does `".*p"`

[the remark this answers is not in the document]
TimB87: fixed with 267f325

900.version-fixes/p.yaml
TimB87: fixed with a538e06

900.version-fixes/perl.yaml
TimB87: fixed with ec28728

900.version-fixes/f.yaml
TimB87: Are Windows repositories generally marked as untrusted? I ask because of "AppGet"

AMDmi3:
> This PR will be marked as a draft to fill it up and have a place to ask about specific rules, if that's ok with you?

Please don't, I don't want to review the same diff multiple times. PR per bunch of fixes is OK.

900.version-fixes/f.yaml
AMDmi3: Yes. Not because they are windows, but because they all fake a lot of versions.

900.version-fixes/python.yaml
```diff
@@ -101,6 +101,7 @@
 - { name: "python:shiboken",           ver: "2.0.0",                 ruleset: kaos,        incorrect: true }
 - { name: "python:simpleparse",        ver: "3.0.0",                 ruleset: sisyphus,    incorrect: true }
 - { name: "python:simpleparse",                                      ruleset: sisyphus,    untrusted: true } # accused of fake 3.0.0
+- { name: "python:sip",                verpat: "5\\..*",                                   incorrect: true }
```
AMDmi3, line 104: I don't see anything incorrect about sip 5.x

TimB87:
Ok, i removed the draft status

900.version-fixes/x.yaml
```diff
@@ -38,6 +38,7 @@
 - { name: xdg-utils,                   verpat: "20[0-9]{6}",                               snapshot: true }
 - { name: xdot,                        ver: "20121124",                                    snapshot: true }
 - { name: xdotool,                     verpat: "20[0-9]{6}.*",                             incorrect: true } # it's 3.YYYYMMDD; XXX: problem
+- { name: xdotool,                     ver: "3.20200618.1",                                incorrect: true } # the tag they pull is from 2016..
```
AMDmi3, line 41: Please limit this with `ruleset: openmandriva` and add `ruleset: openmandriva, untrusted: true` to ignore possible future incorrect versions.

900.version-fixes/x.yaml
```diff
@@ -107,7 +108,9 @@
 - { name: xrdp,                                                      ruleset: sisyphus,    untrusted: true } # Accused of fake 0.9.12.1
 - { name: xroach,                      verge: "12",                  ruleset: [opensuse,rpmsphere], incorrect: true }
 - { name: xrootconsole,                verlonger: 2,                 ruleset: void,        debianism: true }
-- { name: xsel,                        verpat: ".*20[0-9]{6}.*",                           snapshot: true }
+- { name: xsel,                        ver: "1.2.0",                                       outdated: true } # 1.2.0 is 12 years old
```
AMDmi3, line 111: Snapshots cannot be compared, and 1.2.0 is the latest official, so please leave as is. There also can't be `noscheme` on a subset of packages.

900.version-fixes/s.yaml
```diff
@@ -236,6 +236,7 @@
 - { name: stp,                         verlonger: 3,                 ruleset: freebsd,     incorrect: true }
 - { name: stp,                         verpat: "[0-9]{4,}",                                incorrect: true }
 - { name: stp,                         verpat: "201[56].*",                                outdated: true }
+- { name: strace,                      verpat: ".*ab.*", devel: true } # the version is technically never defined and rather incorrect than devel
```
AMDmi3, line 239: Better mark as incorrect + additional untrusted rule for sisyphus and fedora

900.version-fixes/p.yaml
```diff
@@ -14,6 +14,7 @@
 - { name: pangox-compat,               verlonger: 3,                 ruleset: rosa,        incorrect: true }
 - { name: paper-icon-theme,            verpat: "20[0-9]{2}.*",                             snapshot: true }
 - { name: paper-icon-theme,            verpat: "2017.*",                                   outdated: true } # 1.5.0 from 2018
+- { name: papirus-libreoffice-theme,   ver: "20170914",                                    devel: true } # that version was never tagged
```
AMDmi3, line 17: devel → incorrect + additional untrusted rule for aosc, please

900.version-fixes/python.yaml
TimB87: While they share the same project name, they don't seem to be the same. Qutebrowser needs both as build dependencies for example

900.version-fixes/p.yaml
TimB87: See fa31325

900.version-fixes/s.yaml
TimB87: See 800b04a

900.version-fixes/x.yaml
TimB87: see 3e039be

900.version-fixes/x.yaml
TimB87: see b218eec

900.version-fixes/python.yaml
AMDmi3: They are definitely different versions of the same project.
https://www.riverbankcomputing.com/hg/sip
https://pypi.org/project/sip/#history
Please remove this rule and this PR can be merged.

TimB87: 0f4f595 done
I might want to mark versions 4.x.y as legacy in a later step then, but we can do that in another round :+1:

Issue #353, opened 2020-03-29
Possible issues with texlive-xetex packages


waldyrious:
Not sure if this is the right repository to report this, but the packages [texlive:xetex](https://repology.org/project/texlive:xetex/information) and [texlive-xetex](https://repology.org/project/texlive-xetex/information) seem to refer to the same project. Is their separation intentional?

In addition to that, I'm also puzzled as to why Ubuntu's [texlive-xetex](https://packages.ubuntu.com/eoan/texlive-xetex) does not show up in either of them.

AMDmi3:
> Is their separation intentional?

It's not intentional, but I don't consider working on properly merging it worth the effort as it doesn't have a clean versioning scheme anyway. The whole texlive ecosystem looks messy in fact. Using CTAN as a reference data source is also [not possible](https://github.com/repology/repology-updater/issues/886) now.

> In addition to that, I'm also puzzled as to why Ubuntu's texlive-xetex does not show up in either of them.

Because it's https://packages.ubuntu.com/source/eoan/texlive-base

waldyrious:
Thanks for the explanation. I'm not sure I understand the last point, though. The `texlive-xetex` page says that it _depends_ on `texlive-base`, but also that `texlive-base` is its source... are these two different types of entities?

AMDmi3:
I suppose source and binary packages are completely different namespaces.

waldyrious:
Ah, I didn't realize the two links pointed to different URLs, my bad. Does repology only cover source packages?

AMDmi3:
Yes, only source for now

waldyrious:
Ok, thanks, that wasn't clear to me until now. I suppose there's nothing actionable in this issue, then? Or could I work on merging the texlive-xetex/texlive:xetex packages myself? If so, please let me know how to proceed.

PR #385, opened 2020-06-30
Add two new aliases to `fonts:gnu-freefont`

waldyrious:
- `font-gnu-freefont`, used by Guix: https://guix.gnu.org/packages/font-gnu-freefont-20120503/
- `font-freefont-ttf`, used by Solus: https://dev.getsol.us/source/font-freefont-ttf/

### DIFF
--- a/800.renames-and-merges/fonts/g.yaml
+++ b/800.renames-and-merges/fonts/g.yaml
@@ -188,6 +188,7 @@
 
 - setname: "fonts:gnu-freefont"
   name:
+    - font-freefont-ttf
     - font-gnu-freefont
     - font-gnu-freefont-ttf
     - fonts-freefont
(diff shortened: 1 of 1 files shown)


waldyrious:
Should I also add `font-freefont-ttf`, [used by the Solus project](https://dev.getsol.us/source/font-freefont-ttf/)? Although that may require an entry in `900.version-fixes/` as well, because of the nonstandard versioning they're using (I assume this would be similar to the issue discussed in #353).

AMDmi3:
> Should I also add font-freefont-ttf, used by the Solus project?

Yes please

> Although that may require an entry in 900.version-fixes/ as well

Not needed since 5.3.12 compares less than 20120503

AMDmi3:
> Should I also add font-freefont-ttf, used by the Solus project?

Yes please

> Although that may require an entry in 900.version-fixes/ as well

Not needed since 5.3.12 compares less than 20120503

waldyrious:
Done @AMDmi3

PR #402, opened 2020-08-20
Add split for rmtrash

waldyrious:
The [macOS software](http://www.nightproductions.net/cli.htm) included in Homebrew and MacPorts is different from the [Linux one](https://github.com/PhrozenByte/rmtrash).

Note that this is similar to the split of `trash` done in 57b334d. @AMDmi3 I wonder if these two should have the suffix `-macos` instead of `-osx`; WDYT?

### DIFF
--- a/850.split-ambiguities/r.yaml
+++ b/850.split-ambiguities/r.yaml
@@ -59,6 +59,10 @@
 - { name: ricochet, ruleset: solus, setname: ricochet-im }
 - { name: ricochet, addflag: unclassified }
 
+- { name: rmtrash, wwwpart: nightproductions, setname: rmtrash-nightproductions }
+- { name: rmtrash, wwwpart: PhrozenByte, setname: rmtrash-phrozenbyte }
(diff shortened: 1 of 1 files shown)


AMDmi3:
It looks like the other rmtrash is designated for OS X too as it's a wrapper around osx trash-put. Not sure what it does in AUR, but pkgsrc and nix are available on osx so it makes sense. Taking this into account, I'd disambiguate it based on vendor (e.g. into rmtrash-phrozenbyte  and rmtrash-nightproductions). In any case, please split it completely, e.g. provide a rule per each project moving it away from ambiguous name and a catch-all `warning` rule, like it's done with `rum`, for example.

waldyrious:
@AMDmi3 thanks for the detailed response and sorry for the delay. I've now updated the PR as you recommended.

AMDmi3:
Please rebase and update warning rule to `addflag: unclassified`

waldyrious:
Done!

PR #405, opened 2020-08-27
split gf

jube:
`gamedev-framework` and `grammatical-framework` both use `gf` as their package name in different repositories. The problem has already been reported [here](https://repology.org/project/gf/report).

In order to be fair, I did not kept the `gf` name for either, I made the names more explicit for both (and `gamedev-framework` already exists in vcpkg and in [repology](https://repology.org/project/gamedev-framework/versions) too).


AMDmi3:
Good, but please add third `addflag: unclassified` rule to make sure nothing remains under `gf`

jube:
Done!

PR #431, opened 2020-11-04
Couple minor fixes for texlive

veprbl:


AMDmi3:
I believe that there should only be single `YYYYMMDD` version scheme supported, as it's the most precise and commonly used. `YYYY.*` are just random, and `YYYY` artifically loses precision:
```
    src = fetchurl {
      urls = [
        "http://ftp.math.utah.edu/pub/tex/historic/systems/texlive/${year}/texlive-${year}0406-source.tar.xz"
              "ftp://tug.ctan.org/pub/tex/historic/systems/texlive/${year}/texlive-${year}0406-source.tar.xz"
      ];
```
`YYYYMMDD` is also the only scheme compatible with vulnerability database.

New rules for merging nix-specific package variants are also not accepted - repositories are required to provide information required to merge packages of the same project.

veprbl:
I will argue that the `YYYYMMDD` scheme is not more precise. There is only one official release of the tarballs for the binaries, so, for example, 2020 directly maps onto 20200406. The packages that go into texmf are rolling release, the distributions have to mirror some snapshot for a single day (you can see for yourself https://texlive.info/tlnet-archive/2020/). So, I suppose, the most reasonable scheme would probably be 2020.20201104 for binaries from 2020 and snapshot of November 4th.

PR #432, opened 2020-11-09
Remove exult from 900.version-fixes

winterheart:
Version 1.6 released, ignore non semantic versions.


AMDmi3:
Thanks! I've tweaked these a bit - we want to mark _all_ snapshots regardless of repository as snapshots, and we can also mark snapshots known to be outdated based on latest release date as such.

PR #444, opened 2020-12-21
899..global: match gentoo snapshot patch releases as ignore

telans:
Matches versions such as `1.0_p20200130` which are actually snapshots post release
Does not match versions such as `11.0.9_p11` which are normal patch releases

https://regex101.com/r/adXgXl/2

### DIFF
--- a/899.version-fixes.global.yaml
+++ b/899.version-fixes.global.yaml
@@ -59,6 +59,7 @@
 - { verpat: ".*r.*", namepat: "php:.*",                              ruleset: freebsd,     devel: true }
 - { verpat: "5\\.[0-9]{2}\\.0",                                      ruleset: kaos_build,  untrusted: true } # accused of faking -rc versions of KDE stuff as release
 - { verpat: "[0-9.]+_git.+",                                         ruleset: alpine,      ignore: true } # snapshots
+- { verpat: ".*_p20[0-9]{6}",                                        ruleset: gentoo,      ignore: true } # snapshots
 
 # nuget is terminally broken and can't handle arbitrary version schemes, so most are garbage
 - {                                                                  ruleset: chocolatey,  untrusted: true }


AMDmi3:
Not needed, similar rule is already there: https://github.com/repology/repology-rules/blob/master/999.postnormalize.yaml#L6

telans:
@AMDmi3 In that case I think the file should be updated to reflect that all `_p` prereleases are suffixed with at least yyyymmdd, and patch releases using (for example) `_p1` should not be marked as untrusted or ignore.

In general with this rule I believe it should be `ignore` rather than `untrusted`, as these are snapshots, not incorrect versions.

AMDmi3:
> patch releases using (for example) _p1 should not be marked as untrusted or ignore

They should. As it can be [seen](https://repology.org/projects/?inrepo=gentoo&problematic=on) _p* which are not YYYYMMDD are fake versions as well. Rare cases where `p` is actually part of an official version are handled correctly given there's corresponding `p_is_patch` rule for a project.

> In general with this rule I believe it should be ignore rather than untrusted, as these are snapshots, not incorrect versions. 

This is true, but the usage of ambiguous `p` makes them closer to incorrect as generic implementation won't be able to compare these against release versions correctly.

telans:
> As it can be [seen](https://repology.org/projects/?inrepo=gentoo&problematic=on) _p* which are not YYYYMMDD are fake versions as well.

While just about all the `_p[0-9]{1}` gentoo packages are not suffixed as a patch release in other repos, the upstream is usually from debian/ubuntu which does frequently re-release packages of the same version with minor changes. Below are the first 5 `_p[0-9]{1}` packages from the link you provided, all of which actually *do* have patch releases.

https://repology.org/project/an/versions
https://packages.debian.org/unstable/games/an

https://repology.org/project/angrydd/versions
https://packages.debian.org/sid/angrydd

https://repology.org/project/apmd/versions
https://packages.debian.org/oldstable/source/apmd

https://repology.org/project/apt-cacher-ng/versions
https://packages.debian.org/unstable/source/apt-cacher-ng

https://repology.org/project/atheme-services/versions
https://github.com/atheme/atheme/releases/tag/v7.2.10-r2

> This is true, but the usage of ambiguous p makes them closer to incorrect as generic implementation won't be able to compare these against release versions correctly.

I'm not sure specific patch releases should even be compared, it would be suitable to just strip the `_p.*` and compare base versions (excluding snapshots). That's what seems to happen with Arch Linux packages that use `-[0-9]` as a sort of patch release I believe.

AMDmi3:
With Debian versions, Gentoo makes _revision_ (which Repology ignores, in both Debian and Gentoo) part of the _version_, and on top of that, this is done in a way not even compatible with Debian itself. There's no way to treat these normally. These can't be stripped either as they are supposed to be part of the official versions, as with, for instance, sudo, openssh and bash.

telans:
I guess then the only way to force a correct `_p` (i.e `0.8_p1` for `0.8.post1`) package out of "untrusted" would be to make PR excluding it. I'd still like to reiterate my request for marking `_p20[0-9]{6}` snapshots as "ignore" even if in general `_p` is ambiguous. Versions matching this are always snapshots. Cheers

AMDmi3:
Ok, done.

PR #450, opened 2021-01-04
Merge arbtt and haskell:arbtt

waldyrious:
These two entries are the same software: https://repology.org/projects/?search=arbtt


AMDmi3:
If it's end-user software and not a module, it should be merged to arbtt

waldyrious:
Oh, right, my bad. I've fixed it now.

PR #456, opened 2021-01-31
Add rule for SReview

yoe:
SReview is packaged on CPAN as SReview, and in Debian as sreview (not libsreview-perl, as it's a program, not a library)

Note though that perl:sreview doesn't currently show up on repology.org. I don't know whether that's a separate thing that needs to be taken care of, or whether it's expected and will resolve itself...?

### DIFF
--- a/800.renames-and-merges/s.yaml
+++ b/800.renames-and-merges/s.yaml
@@ -244,6 +244,7 @@
 - { setname: squirrel-sql,             name: squirrelsqlstandard, addflavor: standard }
 - { setname: squirrelmail-$1,          namepat: "squirrelmail-(.*)-plugin", ruleset: freebsd }
 - { setname: sratom,                   name: libsratom }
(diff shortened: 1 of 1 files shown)


AMDmi3:
> Note though that perl:sreview doesn't currently show up on repology.org

This is intended, we don't want to pollute projects list with all modules from cpan/pypi/rubygems/etc which are not packaged anywhere.

yoe:
On Sun, Jan 31, 2021 at 04:56:27AM -0800, Dmitry Marakasov wrote:
>      Note though that perl:sreview doesn't currently show up on repology.org
> 
>    This is intended, we don't want to pollute projects list with all modules
>    from cpan/pypi/rubygems/etc which are not packaged anywhere.

Okay. I had guessed as much, but wasn't 100% sure, so just wanted to point it out.

-- 
To the thief who stole my anti-depressants: I hope you're happy

  -- seen somewhere on the Internet on a photo of a billboard

PR #459, opened 2021-02-05
merge aerial theme from gentoo-guru and arch-aur

epsilon-0:
Signed-off-by: Aisha Tammy <floss@bsd.ac>


AMDmi3:
Thanks! I've also added noscheme rule as it doesn't seem to have official releases.

epsilon-0:
Yep, thanks a lot :heart:

PR #468, opened 2021-03-19
ntfs-3g: normalize the incorrect AR scheme to AR

Artoria2e5:
It feels a bit more useful to just normalize the scheme instead of marking it as wrong.

### DIFF
--- a/900.version-fixes/n.yaml
+++ b/900.version-fixes/n.yaml
@@ -103,7 +103,7 @@
 - { name: nrg2iso,                     ver: "0.5",                   ruleset: opensuse,    incorrect: true }
 - { name: nrg2iso,                                                   ruleset: opensuse,    untrusted: true } # accused of fake 0.5
 - { name: nss-ldap,                    verpat: "1\\.([0-9]{3,})",                          setver: "$1" } # FreeBSD garbage
-- { name: ntfs-3g,                     verpat: "(\\.[0-9]+){3}(\\.[0-9]+)",                setver: "$1AR$2" } # incorrect scheme for ARs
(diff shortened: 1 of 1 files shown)


AMDmi3:
Incorrect versions must be fixed in repository, we don't fix these here.

Artoria2e5:
The thing is it might be hard to define what's incorrect. It's totally possible that some distro has a rigid version format, and this is the only way to express that info.

Or in other cases, it can actually be simply more specific than what repology wants. An example is the CUDA package at https://repology.org/project/cuda/related. The upstream actually releases versions with the driver version included, which some distro choose to retain and repology decided to mark as problematic.

What I am saying is that there's a point in normalizing unambiguous versions expressed differently. There's a conversation to be had.

AMDmi3:
Normalizing incorrect versions in Repology may look like an easy solution, but is not beneficial in long term. It won't fix incorrect versions in repositories, and will remove motivation to do so. It would hurt all other consumers of these incorrect repository versions, for instance there may be other new upstream version reporters or security-related tools which would fail to process these versions correctly. We want to improve ecosystem in general, after all. On Repology side, it would just blow up the ruleset and require extra maintenance.

> It's totally possible that some distro has a rigid version format, and this is the only way to express that info.

True, but there are many solutions to this. Repository tools may be improved to support more flexible version formats, we even have a [library](https://github.com/repology/libversion) for it now which has proven itself viable on lots of projects and repositories which Repology covers. And you can see that all other repositories do not have problems with these AR versions.

Alternatively, you can contact upstream and ask them to use more compatible versioning scheme. I've succeeded with that multiple times (iputils, gmic-qt, wizznic, libnumbertext, binaryen).

> Or in other cases, it can actually be simply more specific than what repology wants

These are quite different cases. As long as that's upstream which provides different versioning schemes, such as X.Y.Z and X.Y.Z.BUILD, Repology supports these as in https://repology.org/project/yacreader/information. The same thing can be used for cuda.

PR #470, opened 2021-03-23
Don't mark latest clojure version as incorrect

jlesquembre:
Latest clojure version (`1.10.3.814`) is marked as incorrect, but that's
the latest released version. Official clojure changelog can be found
here:
https://github.com/clojure/brew-install/blob/1.10.3/CHANGELOG.md

There is an official download for `1.10.3.814`:
https://download.clojure.org/install/linux-install-1.10.3.814.sh

But not for `1.10.3`:
https://download.clojure.org/install/linux-install-1.10.3.sh

Before `1.9.0`, clojure versions followed this pattern: `x.y.z`, but since
`1.9.0`, the pattern is `x.y.z.nnn`.

### DIFF
--- a/900.version-fixes/c.yaml
+++ b/900.version-fixes/c.yaml
@@ -152,7 +152,7 @@
 - { name: clisp,                       ver: "2.50",                                        untrusted: true } # very likely fake, see reports
 - { name: clisp,                       vereq: "2.49.93",             ruleset: [arch,freebsd,fedora,manjaro,ravenports], incorrect: true }
 - { name: clisp,                                                     ruleset: [arch,freebsd,fedora,manjaro,ravenports], untrusted: true } # accused of fake 2.49.93
-- { name: clojure,                     verpat: "[0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]{3,}",     incorrect: true } # homebrew and nix garbage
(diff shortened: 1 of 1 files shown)


AMDmi3:
There are no 4 component versions in tags: https://github.com/clojure/clojure/tags

jlesquembre:
The 4 component version is here:
https://github.com/clojure/brew-install/releases
Which is the clojure lang + cli tools

AMDmi3:
> https://github.com/clojure/brew-install/releases

So it's a brew specific thing.

jlesquembre:
> > https://github.com/clojure/brew-install/releases
> 
> So it's a brew specific thing.

not really, for linux distros is the same versioning. I'm not a brew user.

It's a bit confusing, because with version 1.9.0 they introduced some CLI tools to interact with the clojure interpreter. The language and the CLI tools are maintained in different repositories, but you want both to use clojure.


You can see the latest version here: https://clojure.org/guides/getting_started#_installation_on_linux

AMDmi3:
Ok, still the submission is not correct as these are not real closure versions and removing the rule would make correct 3 comp versions outdated. You could change `incorrect` to `altver` instead, this way both 3 component and 4 component versions would be handled fine.

jlesquembre:
ok, I pushed new changes, is that what you meant?

AMDmi3:
Yes, this should be good, thanks!

jlesquembre:
Thanks to you for all your work on repology, it's a very useful project :)

PR #472, opened 2021-03-26
Merge names for libmt32emu

dreamer:
Repository packagers are confused, as the lib lives within the repo of
the munt project.


AMDmi3:
It looks like these all should be be merged into `munt` instead with `flavor: mt32emu`.
It would be nice to merge `munt-alsadrv` and `munt-mt32emu-alsadrv` into munt along the way (with `flavor: alsadrv`).

dreamer:
Munt is an application using the mt32emu lib, with a separate directory / cmake project. It does share the repo with mt32emu and mt32emu-alsadrv though (historical artifact of being stored in the SVN in the past).

AMDmi3:
> with a separate directory / cmake project

Not really a reason to treat it as a separate project, but as I see now they don't share a release scheme, so the submission is valid. alsadrv does not seem to have separate releases though so it still needs to be merged.

PR #479, opened 2021-05-12
Disambiguate mcedit-unified

waldyrious:
There are currently two packages titled simply [mcedit](https://repology.org/project/mcedit), which should be merged into [mcedit-unified](https://repology.org/project/mcedit-unified).

Not sure if the fact that there will be two AUR packages named `mcedit-unified` will be an issue; both _do_ correspond to the same project:

- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mcedit-git
- https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mcedit-unified

### DIFF
--- a/850.split-ambiguities/m.yaml
+++ b/850.split-ambiguities/m.yaml
@@ -54,6 +54,9 @@
 - { name: mc, wwwpart: minio, setname: minio-mc }
 - { name: mc, wwwpart: fetk, setname: mc-manifold-code }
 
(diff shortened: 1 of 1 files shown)


waldyrious:
Also, should I add a rule to disabmiguate [mcedit-editor](https://repology.org/project/mcedit-editor/versions) as well?

AMDmi3:
This looks good, thanks. mcedit-editor does not clash with this so may be left be, although it could be removed since it's a alt linux specific dud.

PR #481, opened 2021-05-13
Fix Chromium-prefixed version for Brave

waldyrious:
Not sure I follow the rationale for the existing Brave rule, btw 🤔


AMDmi3:
This should set to `incorrect` instead of normalizing, for this is a fake version.
Older rule is outdated and may be removed.

waldyrious:
Ok. Should I still match it with `verpat`, or use `ruleset: homebrew`?

waldyrious:
I guess `vercomps` should be enough since we don't need to capture the version, and is more generic than singling out a specific family.

PR #483, opened 2021-05-20
Fix latest kitty version

kovidgoyal:
Some repos such as Funtoo and AUR use additional build numbers after the
main kitty version. This causes the latest version to be incorrectly
detected. So ignore anything after the first three numbers.


AMDmi3:
Please use `verpat: ".+20[0-9]{6}", snapshot: true` for this.

kovidgoyal:
I dont follow, the current bad version in funtoo is 0.20.3.202 the correct version is 0.20.3 how would your suggested verpat filter this?

kovidgoyal:
Never mind i see its actually a date

kovidgoyal:
changed as requested.

PR #486, opened 2021-05-25
Merge dua

austinbutler:
https://repology.org/project/dua/versions
https://repology.org/project/dua-cli/versions


AMDmi3:
Good, but you probably also want to merge "rust:dua-cli". Note that I'll merge this a bit later, for this is a testcase for repology/repology-updater#1153.

austinbutler:
Do you mean in addition to `dua` or instead of?

AMDmi3:
In addition to.

PR #487, opened 2021-06-07
merge superslicer-prerelease to superslicer

Lukas1818:


Lukas1818:
I am not sure what the difference between `weak_devel` and `devel` is.

AMDmi3:
Please also add `nolegacy: true` and squash into a single commit.

> I am not sure what the difference between `weak_devel` and `devel` is.

`devel` is for cases where we known for sure that this is a correct official unstable version (for instance, when this information comes from trusted source such as Wikidata). It marks all packages with equally compared version as devel status.
`weak_devel` is for cases where the package name suggests that the version is unstable, but there's no guarantee it doesn't contain unofficial version (snapshot), isn't outdated by newer _stable_ version, or isn't updated to a stable version itself (to avoid outdated package in the repository). So it's not trustworthy to affect other packages' statuses and it's marked as ignored status.

AMDmi3:
Thanks!

Lukas1818:
I had not squash it yet.

PR #492, opened 2021-07-02
Merge xone

mkopec:
Merge `xone` with `xone-dkms`, which is the name we use in the AUR: https://aur.archlinux.org/packages/xone-dkms-git/

### DIFF
--- a/800.renames-and-merges/x.yaml
+++ b/800.renames-and-merges/x.yaml
@@ -82,6 +82,7 @@
 - { setname: xmonad,                   name: "haskell:xmonad" }
 - { setname: xmoto,                    name: x-moto }
 - { setname: xmrig,                    name: xmrig-donateless, addflavor: donateless }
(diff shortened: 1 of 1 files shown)


AMDmi3:
This doesn't make sense as there's no `xone` project. `xpadneo` merged.

kakra:
There should also be a xone project, it's just a few days old, probably not picked up by crawlers yet, and maybe only a few distributions. It's part of the Discord family for xpadneo, xow and xone.

mkopec:
I think there are just 2 packaged versions of `xone` at the moment: The AUR packages `xone-dkms{,-git}` that I maintain and a Fedora Copr package: https://copr.fedorainfracloud.org/coprs/sentry/xone/. That's probably why the project doesn't exist on repology yet.

AMDmi3:
Okay, lets merge it for the case where it's added by other repos as `xone`.

mkopec:
Thank you!

PR #494, opened 2021-07-09
Add devel rule for darling-dmg

Luflosi:
The version of darling-dmg in Nixpkgs currently has the version 1.0.4+git20200427, which is using a git revision without a corresponding git tag from the upstream repo, so mark it as a development version.

FYI: this is my first contribution to this repo.


AMDmi3:
This is not quite correct. First, it needs to be merged with `fusefs:darling-dmg`. Next, `devel` is for official unstable versions, snapshots should be `ignore`d, and YYYYMMDD versions are fake and should be marked as `incorrect`. Note that as a result, no repo currently contain valid version.

Luflosi:
Thank you for explaining and for fixing it yourself!

PR #495, opened 2021-07-10
Ignore snapshot release of an ucarp fork

oxzi:
There is an unmaintained ucarp fork out there, from which a snapshot release is used by DPorts and FreeBSD Ports. This change ignores this version.


AMDmi3:
Let's instead use `verpat: ".+20[0-9]{6}` so this pattern is still ignored after update or if it appears in another repo.

oxzi:
Thanks, I have changed the version to your regular expression.

PR #496, opened 2021-07-10
Add noscheme for apfsprogs

Luflosi:
There are no official releases, git tags or similar. `apfsck -v` and `mkapfs -v` print 0.1 as the version and the man pages also show 0.1 but this has been the case since at least 2019: https://github.com/linux-apfs/apfsprogs/commit/db9d5f312fb59dbac118dad948ed5df6c0a9b6ce. According to the commit message "We've been using '0.1' in the man pages for while, so continue with that number until we reach an actual release.".


AMDmi3:
This is indeed not expected behavior, but I don't plan to investigate this now unless since no package is mislabeled outdated.

Luflosi:
It seems to work now: https://repology.org/project/apfsprogs/versions. No idea why but I'm not complaining.

PR #497, opened 2021-07-11
Distinct between Claws Mail version three and four

oxzi:
With its latest release[0], Claws Mail has two versions, 3.18.0 and
4.0.0. The major version three release branch continues depending on
GTK+ 2, while major version four uses GTK+ 3 instead.

This commit separates those two releases by renaming the now older one,
major version three, to claws-mail-gtk2, as it is expected to become
deprecated on the long run.

[0] https://claws-mail.org/NEWS

### DIFF
--- a/950.split-branches.yaml
+++ b/950.split-branches.yaml
@@ -45,6 +45,8 @@
 
 - { name: clanlib,                                        setbranchcomps: 1 }
 
+- { name: claws-mail,                                     setbranchcomps: 1 }
+
 - { name: clutter-gst,                                    setbranchcomps: 1 }
 
 - { name: compiz,                                         setbranchcomps: 2 }


AMDmi3:
We don't split projects based on version. You can add a rule to 950 ruleset to improve handling of legacy versions.

AMDmi3:
Again, you cannot use setname. You can use `setbranchcomps: 1`

oxzi:
Pardon, I failed to understand the effect of `setbranchcomps`. Could you point me to the documentation? The code within the `repology-updated` has not helped me, https://github.com/repology/repology-updater/blob/dffa72da6e7dbb02cc023728606dbeb79b65d658/repology/transformer/rule.py#L681-L687.
Thanks for all your support!

oxzi:
Again, I have updated the commit to match the `setbranchcomps` rule.
The second force push was to alter the commit message.

Even as I am beginning to understanding the effect of `setbranchcomp` due to your [tests](https://github.com/repology/repology-updater/blob/fe40d9a5849f712b6fa7ca63832159e688e821d9/repology/test/test_transformer.py#L138), I simply do not see its outcome in the web interface. The `antlr` project, for example, is also configured wit `setbranchcomps: 1`. However, on the [web page](https://repology.org/project/antlr/information) only its latest major version release is marked as up to date.

Would you please elaborate? Please excuse my questions and thanks a lot!

AMDmi3:
By default, if a repo contains 4.x and 3.x versions, all 3.x would have `legacy` status. With different branch set for 3.x (which is the effect of `setbranchcomps`) 3.x versions have additional evaluation resulting in non-latest 3.x having `outdated` status instead of `legacy`. This is probably not related to your original intention, but a right thing to do in this case.

As a result, repos only having 3.x will be outdated (as they lack latest 4.x), in repos having both 3.x and 4.x non-latest versions in each branch will be labeled outdated. In repo having latest 4.x and 3.x these will be labeled `latest` and `legacy` correspondingly

oxzi:
Indeed, this was not my initial intention, but it seems the sane thing to do. Thanks for your explanation.

PR #498, opened 2021-07-12
Merge qdarkstyle

JeremyRand:
The Arch package for `python:qdarkstyle` and the Debian Testing package for `qdarkstyle` both list https://github.com/ColinDuquesnoy/QDarkStyleSheet as upstream.


AMDmi3:
Sorry, we don't merge misnamed packages. Python module package names expected to have corresponding prefix.

PR #499, opened 2021-07-12
Merge aiorpcx

JeremyRand:
The Arch package for `python:aiorpcx` and the Debian Testing package for `aiorpcx` both list https://github.com/kyuupichan/aiorpcX/ as upstream.


AMDmi3:
Sorry, we don't merge misnamed packages. Python module package names expected to have corresponding prefix.

jbicha:
@AMDmi3 I don't understand your position here. How is this case distinguishable from the other similar entries in the python rename file?

Distros have lots of badly named packages but people expect to find all the packages for the same project on one page.

AMDmi3:
Other entries do not matter, each new rule is considered individually.

Badly named packages should be fixed in distros, not here. There's no possibility of or sense in introducing and maintaining rule for each of massively misnamed package groups, such as python packages in Debian and Slackbuilds. New repositories with such problems are not allowed, for instance. There may be individual exceptions, such as to make a vulnerability visible.

PR #503, opened 2021-08-02
Merge knot3 into knot

vcunat:
It's the name used in FreeBSD ports now.


AMDmi3:
Let's maybe merge `knot3-lib` as well?

vcunat:
It shouldn't hurt, though I haven't noticed `knot2-lib` or `knot3-lib` anywhere.  Maybe I didn't look right.

vcunat:
But I can't see why not.

PR #509, opened 2021-08-16
Merge openstack client libraries

angustrau:
Merge all of the openstackclient CLI plugins with their alternative package names


AMDmi3:
Please merge the other way, to match pypi package names.

AMDmi3:
After some though, I think I'll reject this. All packages which are not `python:python-*` are in fact misnamed, and some clash with unrelated packages (such as `troveclient`) because of that. We don't merge misnamed packages, these need to be fixed in repositories.

PR #518, opened 2021-09-22
Split ansible

xfix:
I made a comment at https://repology.org/project/ansible/report, but I suppose it's barely understandable, so I figured I would upload code explaining what I mean. I don't know if that works, the intent is to make:

- Packages called `ansible-base` or `ansible-core` should be put into `ansible-core`.
- Packages called `ansible` from OpenPKG, Solus, Wikidata should be put into `ansible-core` because they actually package `ansible-core` (well, okay, Wikidata doesn't package anything, but still, they have `ansible-core` version, not `ansible` version).

### DIFF
--- a/800.renames-and-merges/a.yaml
+++ b/800.renames-and-merges/a.yaml
@@ -127,8 +127,8 @@
 - { setname: anki,                     name: [anki12,anki20,"python:anki2"] }
 - { setname: annex,                    name: annex-conquer-the-world }
 - { setname: ansible,                  namepat: "(?:python:)?ansible[0-9]*" }
-- { setname: ansible,                  name: [ansible-base, "python:ansible-base" ] } # XXX: there's a mess in ansible releases (ansible vs. core vs. base and 2.x vs 3.x)
-- { setname: ansible,                  name: [ansible-core, "python:ansible-core" ] } # would be nice of someone with knowledge to sort this out
+- { setname: ansible-core,             name: ansible, ruleset: [openpkg, solus] }
+- { setname: ansible-core,             name: [ansible-base, "python:ansible-base"] }
 - { setname: ansible-lint,             name: "python:ansible-lint" }
 - { setname: ansible-runner,           name: "python:ansible-runner" }
 - { setname: ant,                      name: ant-junit5, addflavor: true }
--- a/950.split-branches.yaml
+++ b/950.split-branches.yaml
@@ -2,9 +2,6 @@
 
 - { name: allegro,           verlt: "4.9",                setname: allegro4 }
 
-- { name: ansible,           verge: "3",                  setname: ansible-collections }
-- { name: ansible,           verlt: "3",                  setbranchcomps: 2 }
-
 - { name: antlr,                                          setbranchcomps: 1 }
 
 - { name: apache,                                         setbranchcomps: 2 }


xfix:
Actually, I can fix data in Wikidata myself, so may as well remove an exception for Wikidata from my pull request.

AMDmi3:
Pasting info from report.
```
Okay, this is a mess. To explain what's going on.

- PyPI's ansible 2.9 and older contain both ansible-base and ansible-collections. There is no package on PyPI called ansible-base.
- PyPI's ansible 2.10 split ansible-base and ansible-collections. PyPI's ansible package bundles both.
- PyPI's ansible 3 contains ansible-base 2.10 and ansible-collections.
- PyPI's ansible 4 has ansible-collections and ansible-base which was renamed to ansible-core 2.11.

I propose a following scheme to deal with the mess that versioning of this package is.

- Any package called ansible-core or ansible-base would be put into ansible-core. ansible-base being an older name for ansible-core.
- OpenPKG, Solus and Wikibase releases should be moved into ansible-core. 2.11 was never a version of ansible (where ansible is a bundle of both ansible-base/ansible-core and ansible-collections). I checked packages that OpenPKG and Solus have, and they package ansible-core only. I suggest specifically moving entire distros into this scheme because that will make it more obvious when they stop using that scheme (ansible-core will suddenly get a new version that doesn't actually exist) as opposed to blacklisting specific versions (which could be way messier to deal with if ansible-core were to get 3.0 release).
- Get rid of ansible-collections package, move packages from it to ansible. No distribution appears to have ansible-collections by themselves (without ansible-base/ansible-core part)

After doing this, the newest version of ansible should be 4.6.0, and newest version of ansible-core should be 2.11.5. 
```

AMDmi3:
- I don't see any rationale behind this. What's wrong with the current scheme?
- Splitting/merging by ruleset is not allowed.

I believe this topic was already risen, but I cannot find the issue. As far as I remember, the only unresolved question was whether projects should be named `ansible`+`ansible-collections` or `ansible-core`+`ansible`. My rationale is that upstream project is called `ansible` (https://github.com/ansible/ansible, https://releases.ansible.com/ansible/) and I'm not even sure what 3.x and 4.x stuff is and where it comes from. If it's decided to rename it to `ansible-core`+`ansible`, there's no need to touch anything beside 950 ruleset.

xfix:
https://github.com/ansible/ansible contains `ansible-core`. I assume the repo is called `ansible` because they didn't want to break already cloned repositories, but admittedly it makes things much more confusing. The image on https://www.ansible.com/blog/announcing-the-community-ansible-3.0.0-package makes it clear the project in `ansible` repository is called Ansible Core.

About `ansible-collections`, as far as I can tell, no distribution actually packages `ansible-collections` by itself.

3.x and 4.x versions come from PyPI, see https://pypi.org/project/ansible/. The `ansible` package is essentially a bundle of `ansible-core` and `ansible-collections` that has its own version number. This makes sense even for Ansible 2.9 and older, as back then `ansible-core` and `ansible-collections` were in a single repo.

AMDmi3:
> 3.x and 4.x versions come from PyPI

I don't understand this. PyPI is just a way to publish packages, there should be a genuine repository with 3.x and 4.x tags somewhere.

Anyway, I'm not against renaming it, you could update the PR to just modify two rules in 950 file.

xfix:
> I don't understand this. PyPI is just a way to publish packages, there should be a genuine repository with 3.x and 4.x tags somewhere.

After searching for a while, I believe https://github.com/ansible-community/antsibull-changelog is used for deploying PyPI packages, and https://github.com/ansible-community/ansible-build-data has specific versions of packages to include.

xfix:
> Anyway, I'm not against renaming it, you could update the PR to just modify two rules in 950 file.

I'm not sure if I understand. 950 seems to be version stuff, and I'm not actually doing anything that checks version numbers. Ansible 2.0 and Ansible 4.0 should be both in `ansible` package on Repology.

My pull request introduces two rules:

```yaml
- { setname: ansible-core,             name: [ansible-base, "python:ansible-base"] }
```

Which is very similar to previously existing rule in this file.

```yaml
- { setname: ansible,                  name: [ansible-core, "python:ansible-core" ] 
```

---

And the second rule I'm honestly not sure about the syntax of: the intent of it was to mark OpenPKG and Solus packages as actually containing `ansible-core` despite its name. I checked packages in those distributions, and they do not contain `ansible-collections` (honestly, I think those distributions are almost certainly confused, `ansible-core` by itself is not very useful, probably nobody noticed because nobody actually uses those packages in those distributions).

```yaml
- { setname: ansible-core,             name: ansible, ruleset: [openpkg, solus] }
```

AMDmi3:
> 950 seems to be version stuff, and I'm not actually doing anything that checks version numbers

Well you should, because there's nothing beside version numbers to check. As mentioned, spitting based on repository names is not allowed. So you'll get `ansible` for >= 3 and `ansible-core` for 2.x, if that's what you propose instead of current `ansible-collections` for >= 3 and `ansible` for 2.x.

xfix:
Maybe I should clarify.

- `ansible` is a package which is usually packaged by distributions as `ansible`.
- `ansible-core` is a package which is usually packaged by distributions as `ansible-base` or `ansible-core`.
- OpenPKG and Solus incorrectly package `ansible-core` as `ansible`. I'm fairly certain this a bug in those distros however.

Version numbers are irrelevant here. They are two distinct packages that should have separate entries on Repology.

AMDmi3:
What's irrelevant is how whoever package whatever. This is messy, as you could've noticed, unreliable and unstable. That is why we don't split based on repository. Instead, we split based on objective criteria such as (usually) URLs. That's likely not possible in this case, but splitting on a version is a workable solution, so that's we do. `ansible-core` getting version 3 is not a problem until it happens.

xfix:
Okay, so ansible versions could be split like this.

- 2.9 and older - `ansible`
- 2.10.0 to 2.10.7 - ???, could be `ansible`, could be `ansible-core`
- 2.10.8 to 2.10.14 - `ansible-core`
- 2.11 - `ansible-core`
- 3.x, 4.x - `ansible`

In this case I guess marking 2.11 as `ansible-core` could be a workable solution, as both Solus and OpenPKG has package claiming to be `ansible` that is actually `ansible-core`.

xfix:
I've no idea how to use version checks, but I guess what could work is marking packages as `ansible-core` if they are either called `ansible-base` or `ansible-core` OR the version is 2.11 or newer, but older than 3.0.

AMDmi3:
In the end, I've merged `ansible-core` and `ansible-base` into `ansible-core` as you've suggested and left `ansible` as-is. Whoever packages ansible-core/base as `ansible` will have it outdated, that's expected bahavior.

PR #550, opened 2021-10-20
Merge sayonara into sayonara-player

FabioLolix:
https://repology.org/projects/?search=sayonara

https://sayonara-player.com/


AMDmi3:
Let's reverse it. Rules in 800 should prefer more generic names, also `sayonara` is more widely used.
It also needs version rules to blacklist `1.7.0_3` and any_is_patch for `1.7.0_stable3`

PR #557, opened 2021-11-08
Fix python:pypcap version by excluding AUR release

oxzi:
The AUR version of the Python pypcap package is a custom fork, only used
within the AUR. This fork is only one patch ahead of the other version
and is not an "official" release.


AMDmi3:
Please add `verlonger: 3` condition. `incorrect` status needs justification.

oxzi:
@AMDmi3: Thanks, I have altered the commit.

PR #567, opened 2021-12-11
AUR legacy: aliza-free, masterpdfeditor-free, typora-free

FabioLolix:
AUR pkgbuilds for proprietary programs which no longer have a free-to-use version, intentionally old version

### DIFF
--- a/220.legacylist.yaml
+++ b/220.legacylist.yaml
@@ -1,9 +1,9 @@
 # vim: tabstop=39 expandtab softtabstop=39
 
-- { name: aliza-free, ruleset: aur, legacy: true }
-- { name: masterpdfeditor-free, ruleset: aur, legacy: true }
-- { name: ruby-mini_magick-4.5, ruleset: aur, legacy: true }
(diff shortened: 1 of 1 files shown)


FabioLolix:
At first sight didn't seen that masterpdfeditor-free was already legacy in 800.renames-and-merges

FabioLolix:
Also typora is already seen as legacy

AMDmi3:
It's not AUR specific and can be handled in a better way, such as by setting `altscheme` on a version range (`verge`). This will lead to these project having two newest versions.

FabioLolix:
> can be handled in a better way, such as by setting altscheme on a version range (verge)

Don't how to set that

I'm closing this since isn't great

AMDmi3:
The required rules (for 900 ruleset) would look like:

`{ name: typora, verge: "1.0", altscheme: true } # paid since 1.0, so allow free version to be preserved`

On a second thought, as long as free versions are no longer released, this (for 220 ruleset) should go too and may be even better:

`{ name: typora, ver: "0.11.18", legacy: true } # last free version`

PR #568, opened 2021-12-11
MLdonkey naming

FabioLolix:
mldonkey is splitted between core and gui for FreeBSD
mldonkey-ed2kad-daemon is a variation on AUR

### DIFF
--- a/800.renames-and-merges/m.yaml
+++ b/800.renames-and-merges/m.yaml
@@ -186,6 +186,7 @@
 - { setname: mixxx,                    name: mixxx-qt5, addflavor: true }
 - { setname: mkcert,                   name: "go:mkcert", addflavor: true }
 - { setname: mksh,                     name: mksh-static, addflavor: true }
+- { setname: mldonkey,                 name: [mldonkey-core,mldonkey-gui,mldonkey-ed2kad-daemon], addflavor: true }
 - { setname: mlt,                      name: [libmlt, mlt-qt4, mlt-qt5,mlt-freeworld,"python:mlt"], addflavor: true }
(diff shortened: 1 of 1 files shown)


800.renames-and-merges/m.yaml
```diff
@@ -186,6 +186,8 @@
 - { setname: mixxx,                    name: mixxx-qt5, addflavor: true }
 - { setname: mkcert,                   name: "go:mkcert", addflavor: true }
 - { setname: mksh,                     name: mksh-static, addflavor: true }
+- { setname: mldonkey,                 name: [mldonkey-core,mldonkey-gui] }
```
AMDmi3, line 189: `addflavor: true` is needed

AMDmi3 (changes requested):

800.renames-and-merges/m.yaml
FabioLolix: For future reference why I need to add addflavor: true for this case?

AMDmi3:
Oh, and it can be merged into a single rule then. Also please squash into a single commit.

FabioLolix:
> Oh, and it can be merged into a single rule then.

Done

> Also please squash into a single commit.

Done
Also I prefer 1 commit per pull request, you can configure github to do it automatically by default

AMDmi3:
> Also I prefer 1 commit per pull request, you can configure github to do it automatically by default

As far as I remember it concatenates commit messages this way producing needless garbage.

FabioLolix:
> As far as I remember it concatenates commit messages this way producing needless garbage.

Maybe no more but I'm not completely sure about that

PR #571, opened 2021-12-11
Correctly parse Guix VCS snapshots as such

nckx:
This should DTRT if https://github.com/repology/repology-updater/issues/920#issuecomment-931445917 is correct.

Because this was missing there are some buggy `ignore`s scattered across the rule set, which this patch can't fix because the problem isn't technical.

### DIFF
--- a/899.version-fixes.global.yaml
+++ b/899.version-fixes.global.yaml
@@ -69,6 +69,7 @@
 - { verpat: "20[0-9]{2}\\-[0-9]+\\-[0-9]+",                          ruleset: spack,       untrusted: true } # snapshots
 - { verpat: "20[0-9]{2}-[0-9]{2}-[0-9]{2}(?:-unstable)?",            ruleset: nix,         untrusted: true }
 - { verpat: ".+20[0-9]{2}-[0-9]{2}-[0-9]{2}.*",                      ruleset: nix,         ignore: true }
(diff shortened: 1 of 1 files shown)


AMDmi3:
This shouldn't be necessary because there's the same thing (with limited scope though) in the parser:

https://github.com/repology/repology-updater/blob/master/repology/parsers/parsers/guix.py#L63

Are there some cases where it doesn't work right?

nckx:
Interesting, I wasn't aware of that special case.

> Are there some cases where it doesn't work right?

I think so but I don't have a good grasp of the entire Repology codebase.

Here, for example, ‘incorrect’ is, well, incorrect: https://repology.org/project/i7z/versions

AMDmi3:
What exactly is incorrect in this case?

nckx:
Yeah, I guess that's my question too.  [This fork](https://github.com/DimitryAndric/i7z/commit/6c342ffe22552de7614a3ccfba6353fb4679f11a) is apparently ‘blessed’ by Repology whilst [this one](https://github.com/afontenot/i7z/tree/v0.28) is not?  Seems arbitrary, is all.

AMDmi3:
The incorrect status is in fact caused by an unrelated rule, neither fork is blessed. In fact 0.28 is much older than 0.27.4, but I guess there's no point in giving preference to any of these as both are inactive, so I'd split them into separate projects.

Are there another questionable cases?

PR #573, opened 2021-12-11
bitbake versioning

FabioLolix:
AUR pkgbuild have a valid version

https://repology.org/project/bitbake/versions

The last stable bitbake version is 1.52.1 (several stable branch exists, 1.52.x is the latest)

https://git.openembedded.org/bitbake


AMDmi3:
For the note, the original rule was correct as pkgbuild referenced a commit prior to 1.34 tag, e.g. not a real 1.34.0 version. Also we don't usually remove untrusted rule but I guess it's ok since the maintainer has changed.

FabioLolix:
> For the note, the original rule was correct as pkgbuild referenced a commit prior to 1.34 tag, e.g. not a real 1.34.0 version.

I don't doubt it was valid in the past

> Also we don't usually remove untrusted rule but I guess it's ok since the maintainer has changed.

The new maintainer is actually me

PR #574, opened 2021-12-11
vim-bitbake noscheme true

FabioLolix:
vim-bitbake don't have fixed releases

https://repology.org/project/vim-bitbake/versions

https://github.com/kergoth/vim-bitbake

### DIFF
--- a/900.version-fixes/v.yaml
+++ b/900.version-fixes/v.yaml
@@ -37,6 +37,7 @@
 - { name: vilistextum,                 verpat: ".*20[0-9]{6}",                             snapshot: true }
 - { name: vim,                         ver: "9999",                                        incorrect: true }
 - { name: vim-airline-themes,                                                              noscheme: true }
+- { name: vim-bitbake,                                                                     noscheme: true }
 - { name: "vim:vim-flake8",            verpat: "20[0-9]{2}.*",                             snapshot: true }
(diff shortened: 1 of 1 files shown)


AMDmi3:
We don't usually add no-op rules. AUR package has rolling status (set for all `*-git` packages in AUR) so it's not compared to Debian versions and can't (incorrectly) outdate them or be outdated by them. Debian versions compare to each other correctly though. So there's no need in this rule unless some package appears which suffers from the version ambiguity.

FabioLolix:
> We don't usually add no-op rules.

What mean no-op?

> it's not compared to Debian versions and can't (incorrectly) outdate them or be outdated by them.

Of this I was sure, for normally versioned programs

> Debian versions compare to each other correctly though. So there's no need in this rule unless some package appears which suffers from the version ambiguity.

Fine
I'll consider this in the future

> AUR package has rolling status (set for all *-git packages in AUR)

At the moment one AUR package, gccxml-git, is intentionally set as outdated

https://repology.org/project/gccxml/versions

Anyway I [asked for its deletion](https://repology.org/project/gccxml/versions) since it old and deprecated

AMDmi3:
> What mean no-op?

A thing that does nothing. E.g. a change with no practical effect.

Issue #558, opened 2021-11-10
Missing crates.io badge


marcospb19:
I'm sorry if I'm missing something, but I think the crates.io package is missing here, and I don't know why.

[![Packaging status](https://repology.org/badge/vertical-allrepos/ouch.svg)](https://repology.org/project/ouch/versions)

This is https://repology.org/badge/vertical-allrepos/ouch.svg, for the project https://github.com/ouch-org/ouch, but there is a version at crates.io: https://crates.io/crates/ouch, why doesn't it appear in the image?

AMDmi3:
Since there's no knowing whether a specific project is an application or module, all crates.io items map to `rust:*` projects (here, [rust:ouch](https://repology.org/project/rust:ouch/versions)) which have to be manually merged with projects without module prefix. You can see crates.io project in "related" tab:

![2021-11-10-154305_maim](https://user-images.githubusercontent.com/474217/141115398-341a2381-3627-4147-aca8-18d8bdb29350.png)

Will merge it now.

marcospb19:
Thanks for the help!

PR #580, opened 2022-01-01
Rename gfold -> rust:gfold

nickgerace:
Fixes a similar issue as https://github.com/repology/repology-rules/issues/558
Closes https://github.com/nickgerace/gfold/issues/160

Thanks for this great project!


nickgerace:
Thank you!

PR #582, opened 2022-01-08
hide the deprecated 1.24.99 release of hledger*

simonmichael:
And any other mistaken release of .99 versions in future.

### DIFF
--- a/900.version-fixes/h.yaml
+++ b/900.version-fixes/h.yaml
@@ -30,6 +30,10 @@
 - { name: hitori,                      ver: "3.24.1",                ruleset: nix,         incorrect: true }
 - { name: hitori,                                                    ruleset: nix,         untrusted: true }
 - { name: hitori,                      verpat: "[0-9]+\\.[0-9]*[13579]\\..*",              devel: true }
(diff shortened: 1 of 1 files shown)


simonmichael:
Background: my 1.24.99 release was an error, and is marked deprecated on Hackage.

simonmichael:
My change looks similar to many existing entries, so I'm a little surprised this was needed, but thanks!

When does this go live, eg affecting badge colours at https://repology.org/projects/?search=hledger ?

AMDmi3:
> My change looks similar to many existing entries, so I'm a little surprised this was needed, but thanks!

I've decided to fix it in the parser, to serve as a reminder to add `preferred_versions` support. Also, this way a correct version will be reported for hackage. You change was in fact good apart from that `hledger-lib` should've been `haskell:hledger-lib`.

> When does this go live, eg affecting badge colours at https://repology.org/projects/?search=hledger ?

In a few hours - I've had to also add a rule for stackage which still exposes recalled version so it'll requre another update cycle.

simonmichael:
Wonderful. Thank you for your work, repology.org is fantastic.

PR #583, opened 2022-01-09
Split xd

Luflosi:
Two repositories contain a package called XD, which is a completely different project compared to the others: https://repology.org/project/xd/versions. The AUR has this package already with the name xd-torrent and it's a good name, so I used it for the other repositories as well.
I kept the `name` upper case because the project name is actually written in upper case, even though I think the code is case insensitive.


AMDmi3:
name should be lowercase here, also this needs corresponding rules for other projects, as well as unclassified rule

Luflosi:
Ok, I changed it to lower case.
I don't know what you mean by "also this needs corresponding rules for other projects, as well as unclassified rule". Can you try to explain it to me please?

AMDmi3:
See how other split rules are composed. For an ambiguous namr there should be a rename rule for each distinct project, and a fallback `-unclassified` rule for packages which don't match any of these.

Luflosi:
Thanks for taking care of it.

PR #584, opened 2022-01-16
Drop odd/even devel rule for deja-dup

mikix:
It now follows a GNOME-style XX.alpha or XX.beta pattern for devel releases.

Source: am the maintainer. But also the [NEWS](https://gitlab.gnome.org/World/deja-dup/-/blob/main/NEWS.md) file.


mikix:
Thanks! And thanks for maintaining repology in general

PR #586, opened 2022-01-17
Add some suffix rules for nix

QuantMint:
Add rules for -unwrapped and -wrapped suffix for nix.


AMDmi3:
Could you please elaborate on what these suffixes mean? Also these rules don't need weak_developed flag.

QuantMint:
> Could you please elaborate on what these suffixes mean? Also these rules don't need weak_developed flag.

You can find an explanation [here](https://github.com/NixOS/nixpkgs/issues/27164) and [here](https://discourse.nixos.org/t/how-to-tell-the-difference-between-wrapped-and-unwrapped-package/6823).
In the end it's the same package with extra overriding capabilities.

QuantMint:
I have just updated the pull request, shortening it and removing "weak_developed".

PR #589, opened 2022-01-27
remove LAStools noscheme rule due to v2.0.0 announcement

StephenWithPH:
Per https://github.com/LAStools/LAStools/releases/tag/v2.0.0:

> Due some requests we start to use the git release feature starting now.
LAStools use a versioning in YYMMDD scheme. This is common for most LAStools users for long time - so we keep this version numbers.
Additionally we will follow the git versioning guidelines for the open source releases. They may also flow into sourcecode next.
Cause we have quite something like a history we start straight away with version 2.0.0.

### DIFF
--- a/900.version-fixes/l.yaml
+++ b/900.version-fixes/l.yaml
@@ -10,7 +10,8 @@
 - { name: langkit,                                                                         noscheme: true }
 - { name: lasem,                       verpat: "[0-9]+\\.[0-9]*[13579]\\..*",              devel: true }
 - { name: lash,                        verpat: "0\\.6\\.0\\.[0-9]{3}.*",                   incorrect: true } # 0.6.0rc2
-- { name: lastools,                                                                        noscheme: true } # 0.6.0rc2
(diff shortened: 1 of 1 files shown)


AMDmi3:
This needs ignore rules for incorrect version schemes.

StephenWithPH:
> This needs ignore and/or sink rules for incorrect version schemes.

I took a stab in 46f094f2ccec1f16ac9fa7ff454d2fb6bf840fdf . I believe this should take care of Chocolatey, nix, and Vcpkg.

[Packaging status](https://repology.org/project/lastools/versions)

AMDmi3:
FYI FreeBSD is banned for providing incorrect versions, but if some other repo provides 2.0.0 it'll be properly handled as latest.

StephenWithPH:
> but if some other repo provides 2.0.0 it'll be properly handled as latest

Thank you for clarifying. I was planning on chasing down the surprising (to me) result of the PR.

FYI- https://github.com/NixOS/nixpkgs/pull/156941

PR #590, opened 2022-02-01
Merge for rofi-wayland

lbonn:
It would be good to have https://aur.archlinux.org/packages/rofi-lbonn-wayland-git/ as a result for AUR in the [rofi-wayland version page](https://repology.org/project/rofi-wayland/versions)

The current project link (https://aur.archlinux.org/packages/rofi-wayland-git) is deprecated and does not work on modern systems but I don't know if marking it as deprecated is needed or will it just be superseded by rofi-lbonn-wayland-git after this change?


AMDmi3:
This will take couple of days to apply because of repology/repology-updater#1222. Also I suggest to use simpler versioning scheme (like `A.B.C.D`) - as you can see, `A.B.C+waylandD` is not compatible with most repositories.

PR #598, opened 2022-02-16
Update python.yaml

ArTourter:
merged nltk to python:nltk
merged xarray to python:xarray

### DIFF
--- a/800.renames-and-merges/python.yaml
+++ b/800.renames-and-merges/python.yaml
@@ -162,6 +162,7 @@
 - { setname: "python:netifaces",       name: netifaces }
 - { setname: "python:networkx",        name: networkx }
 - { setname: "python:networkx",        namepat: "python:networkx[0-9.-]+" }
+- { setname: "python:nltk",            name: nltk }
 - { setname: "python:notebook",        name: ["python:jupyter-notebook", jupyter-notebook] }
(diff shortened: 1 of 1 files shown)


AMDmi3:
Python modules are required to be prefixed in repositories accordingly, so these can me merged automatically. We do not accept manual merges.

ArTourter:
Thanks for the response, I was only trying to emulate the behaviour I see with other packages.

for example, the xarray packages in the slackbuilds.org repository is located in the python/ directory like many of the python packages in this repo. but somehow this does not get given the python: prefix.

other packages from the same repos placed in the gis/ folder, such as shapely, cartopy, pyproj, get properly identified by the site and given the python: prefix.

I will move the nltk package from libraries to python at the next release, but we tend to keep the name the same as they do upstream.

Does that make sense?

AMDmi3:
Not really.
- Repology cannot rely on category here as python/ contains packages which are not python modules, such as python itself
- Moving packages between categories just for Repology sake is bad, we strongly discourage that
- Usually Repology relies on package name prefix, but for some reason most python modules are not prefixed in slackbuilds (while perl, ruby and haskell modules are). I suggest to go this way, as at the very least it improves consistency in slackbuilds.

PR #599, opened 2022-03-22
merges for ory-hydra and vaultwarden-web

epsilon-0:


800.renames-and-merges/o.yaml
```diff
@@ -163,6 +163,7 @@
 - { setname: orocos-kdl,               name: liborocos-kdl }
 - { setname: orocos-kdl,               name: [orocos-kdl-python,orocos-kdl-python2,"python:orocos-kdl","python:python-orocos-kdl"], addflavor: python }
 - { setname: ortp,                     namepat: "(?:lib)?ortp[0-9.-]*" }
+- { setname: ory-hydra,                name: hydra-openid }
```
AMDmi3, line 166: This rule does nothing and is not needed.

AMDmi3 (changes requested):

800.renames-and-merges/o.yaml
epsilon-0: dropped.

PR #602, opened 2022-03-30
Split yass

tibequadorian:


850.split-ambiguities/y.yaml
```diff
@@ -6,6 +6,10 @@
 - { name: yali, wwwpart: pisi, setname: yali-installer }
 - { name: yali, addflag: unclassified }
 
+- { name: yass, wwwpart: kokkinizita, setname: yet-another-scrolling-scope }
```
AMDmi3, line 9: I try to follow these rules when picking unique names:
- keep the original name as a prefix for searchability
- keep the resulting name short for convenience, if possible
- try to convey the topic of the software for clarity, if possible

So IMO it'd be better to use `yass-scrolling-scope` and `yass-genomic-similarity` here.

850.split-ambiguities/y.yaml
```diff
@@ -6,6 +6,10 @@
...
+- { name: yass, wwwpart: bioinfo, setname: yet-another-similarity-searcher }
+- { name: yass, wwwpart: yass-along, setname: yass-karaoke }
```
AMDmi3, line 11: Every block of split rules must end with `addflag: unclassified` rule.

AMDmi3 (changes requested):

850.split-ambiguities/y.yaml
tibequadorian: Ah missed that

850.split-ambiguities/y.yaml
tibequadorian: Good suggestion. I changed them.

AMDmi3 (approved):

AMDmi3:
Thanks!

PR #604, opened 2022-03-31
Update wwwpart of mongoose-web-server

subnut:
Latest URL is mongoose.ws


subnut:
@AMDmi3 Why isn't the change shown in https://repology.org? Why is it still outdated?

PR #605, opened 2022-04-15
Remove obsolete noscheme for nextpnr

Luflosi:
nextpnr started creating releases a couple monts ago and released three versions so far: https://github.com/YosysHQ/nextpnr/releases.


AMDmi3:
This lacks rules to sink or ignore older versions which compare greater than 0.3

Luflosi:
Thanks!

PR #606, opened 2022-04-22
Merge gnome:manager into extension-manager

mjakeman:
Hi,

[extension-manager](https://repology.org/project/extension-manager/versions) is packaged as [gnome-shell-extension-manager](https://repology.org/project/gnome:manager/versions) on Debian and its derivatives.

This merges `gnome:manager` into the more widely used name `extension-manager`.


AMDmi3:
Thank you! Fixed the other way, this should've not been merged into gnome:manager in the first place, as it's not a gnome extension itself.

PR #619, opened 2022-05-28
Split the system monitors (GNOME Shell extensions)

FranklinYu:
GNOME Shell extension store URL:

- https://extensions.gnome.org/extension/120/system-monitor/
- https://extensions.gnome.org/extension/1064/system-monitor/

They served the same purpose, so took the same name. Probably developed independently. As we can tell from the extension identifier number, the 120 one is earlier (and also more popular), so I think it should take the name. I also assumed that I should keep using `gnome:` prefix for shell extensions.

### DIFF
--- a/850.split-ambiguities/g.yaml
+++ b/850.split-ambiguities/g.yaml
@@ -110,6 +110,7 @@
 - { name: gnat-util, ruleset: [freebsd,pkgsrc,ravenports], setname: gnat-util-gcc }
 - { name: gnat-util, addflag: unclassified }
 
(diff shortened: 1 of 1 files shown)


AMDmi3:
The policy is that all projects must be split to new names. There should be a rule for another project, and there should be an unclassified rule for unmatched projects.

FranklinYu:
Thank you for quick response. Nevertheless, I didn’t understand the policy very well: did you mean that I should add an explicit rule for the extension No. 120, instead of setting it as the fallback? Or rather, did you mean that it should not use the name `gnome:system-monitor` either?

AMDmi3:
Both. Each project should be moved to a new unique name.

FranklinYu:
Sorry for forgetting this PR. Thanks a lot for fixing it!

PR #620, opened 2022-06-02
Add version rule for perl:text-xslate

ikesyo:
https://metacpan.org/dist/Text-Xslate


AMDmi3:
What's the point of this rule?

ikesyo:
Sorry for the confusion. Currently https://repology.org/project/perl:text-xslate/versions lists many version-different packages for MetaCPAN. I think `v3.5.9` should be the latest and only package for MetaCPAN. Marking development releases like `1.9999_01` may be meaningless. What is the correct way to filter out the dev version packages?

AMDmi3:
This cannot be fixed by a rule, the problem is in version on metacpan, and it should be fixed there. Note that it has used correct versioning up to 3.4.0.

PR #624, opened 2022-06-25
Merge further packages into maui-nota

alpabrz:
+ If a distribution refers to MauiKit's GitHub Mirrors (e.g. KDE Neon
Unstable does this), Nota should be now matched correctly.
+ OpenMandriva 4.2 does not specify an URL, therefore this entry
needs to be manually moved to the maui-nota project.

I ran `make check`. This should move those two packages in nota-unclassified to their correct project. Hope I did everything alright and feel free to remove that commit message if it's too noisy.

Thank you for maintaining such a valuable project :pray:


alpabrz:
Ah ok thank you for the clarification! Didn't know that.

PR #625, opened 2022-06-29
Merge calindori

alpabrz:
The Alt Linux packages `kde5-calindori` are the same project as `calindori`.
Hope I didn't oversee anything again^^


AMDmi3:
I've added it into a more suitable place.

alpabrz:
Thank you, I should skim through the files better. I'm a rookie, so this feedback really helps :+1:

PR #628, opened 2022-07-09
Split adl

WeebSorceress:
https://repology.org/project/adl/report


WeebSorceress:
Not sure why my commit appeared as Unverified (it's the same key `8E1DE45BD51CC084`) on that first try, that's why I "force pushed".

AMDmi3:
Sorry, I've missed this PR. `adl` was split.

PR #638, opened 2022-08-25
Make a merge rule for Sugarjar

jaymzh:
Created a "report" on this, but then found out I could just submit
a PR!


AMDmi3:
Shouldn't it be `ruby:sugarjar`?

jaymzh:
Yes it should, good catch!

PR #641, opened 2022-09-14
various jitsi naming merges

epsilon-0:
jitsi stuff is very fragmented across different os repos, this tries to give it some semblance of sanity.

To keep things clear, the merged names are the ones from the official deb packages at http://download.jitsi.org/stable/


800.renames-and-merges/j.yaml
```diff
@@ -38,6 +38,12 @@
 - { setname: jetty,                    name: [jetty-minimal, jetty-websocket], addflavor: true }
 - { setname: jfrog-cli,                name: jfrog }
 - { setname: jira-cli,                 name: [jira-cli-go, "python:jira-cli"] }
+- { setname: jitsi,                    name: "jitsi-nightly" }
+- { setname: jitsi-meet,               name: "jitsi-meet-nightly" }
```
AMDmi3, line 42: *-nightly merges require `weak_devel: true, nolegacy: true` flags as well

800.renames-and-merges/j.yaml
```diff
@@ -38,6 +38,12 @@
...
+- { setname: jitsi-meet-electron,      name: "datcom-jitsi-meet-electron" }
```
AMDmi3, line 43: This doesn't look correct as it's a fork

AMDmi3 (changes requested):

800.renames-and-merges/j.yaml
epsilon-0: Ah, your correct. Sorry about this one.

800.renames-and-merges/j.yaml
epsilon-0: done.

epsilon-0:
squashed into one.

AMDmi3 (approved):

PR #644, opened 2022-10-18
Disambiguation for `mdl`

jaymzh:
Disabmiguate `markdownlint` vs `markdownless`

Signed-off-by: Phil Dibowitz <phil@ipom.com>

### DIFF
--- a/800.renames-and-merges/m.yaml
+++ b/800.renames-and-merges/m.yaml
@@ -132,6 +132,8 @@
 - { setname: mdformat,                 name: "python:$0" }
 - { setname: mdk,                      name: gnu-mdk }
 - { setname: mdk,                      name: mdk-doc, addflavor: true }
+# All the MDLs get disambiguated in the 850 ruleset
+- { setname: mdl,                      name: "ruby:mdl" }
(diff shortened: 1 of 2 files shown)


AMDmi3:
Please use `mdl-markdownlint` and `mdl-markdownless` to improve searchability. Also could you please merge "ruby:mdl" into "mdl" (800 ruleset, to avoid duplicate projects for markdownlint. After being merged into `mdl` these packages will be subject to these split rules and will end up in `mdl-markdownlint` as well.

jaymzh:
I'm happy to do that but I don't think most people would search `mdl`. In fact, it took me a while to figure out why I couldn't find it, even GH Actions made around it call it `markdownlint` https://github.com/actionshub/markdownlint

Also, one other thing to consider is there are two `markdownlint`s - the Ruby one (https://github.com/markdownlint/markdownlint/) and also the JS one (https://github.com/DavidAnson/markdownlint), not to mention two different CLIs written around the JS one (https://github.com/igorshubovych/markdownlint-cli and https://github.com/DavidAnson/markdownlint-cli2). So we should make sure none of those are ambiguous.

jaymzh:
OK that update does everything you asked, but also takes into account David's nodeJS version of markdownlint.

PR #653, opened 2022-11-05
Add elpa-php-mode to emacs:php-mode

zonuexe:
Emacs php-mode is now provided in Debian as [elpa-php-mode](https://packages.debian.org/stable/elpa-php-mode). refs https://github.com/emacs-php/php-mode/issues/430

Emacs Lisp packages now tend to be `elpa-` prefixed on Debian.


AMDmi3:
This is not correct and not needed. Repology uses debian source package names.

zonuexe:
OK, thanks.

PR #654, opened 2022-11-14
Translate old AdaCore version scheme to new one for gprbuild and gnatcoll-*

sternenseemann:
Relevant bits from the commit messages:

---

AdaCore seems to have switched version schemes in a way: They used
publish the GNATCOLL libraries with a year as the version number.
Starting with 2021 they seem to use the last two digits of the
year followed by two version parts which have all been zero so far.
This can be seen from the GitHub release pages:

* https://github.com/AdaCore/gnatcoll-core/releases/
* https://github.com/AdaCore/gnatcoll-db/releases/
* https://github.com/AdaCore/gnatcoll-bindings/releases/

The GNATCOLL docs seem to use an YY.0w scheme I can't quite
figure out, but this is unused by any package repository, e.g.

* https://docs.adacore.com/gnatcoll-docs/
* https://docs.adacore.com/live/wave/gnatcoll-core/html/gnatcoll-core_ug/index.html
* https://docs.adacore.com/live/wave/gnatcoll-iconv/html/gnatcoll-iconv_ug/index.html
* https://docs.adacore.com/live/wave/gnatcoll-syslog/html/gnatcoll-syslog_ug/index.html
* https://docs.adacore.com/live/wave/gnatcoll-readline/html/gnatcoll-readline_ug/index.html

---

For gprbuild it's the same story as with gnatcoll-*: AdaCore switched
to using YY.0.0 version strings over a 20YY year number in 2021:
https://github.com/AdaCore/gprbuild/releases

The alternative version number of YY.0w is only used in the docs,
but in no package repository at the moment:
https://docs.adacore.com/gprbuild-docs/html/gprbuild_ug.html

### DIFF
--- a/900.version-fixes/g.yaml
+++ b/900.version-fixes/g.yaml
@@ -141,8 +141,9 @@
 - { name: gnash,                       ver: "0.8.11",                ruleset: debuntu,     ignore: true } # not released yet
 - { name: gnat-programming-studio,                                                         noscheme: true }
 - { name: gnat-util-gcc,                                                                   noscheme: true }
-- { name: gnatcoll-bindings,                                                               noscheme: true }
-- { name: gnatcoll-db,                                                                     noscheme: true }
+- { name: gnatcoll-bindings,           verpat: "20[0-9]{2}",                               sink: true }
+- { name: gnatcoll-core,               verpat: "20[0-9]{2}",                               sink: true }
+- { name: gnatcoll-db,                 verpat: "20[0-9]{2}",                               sink: true }
 - { name: gnome-2048,                  verpat: "[0-9]+\\.[0-9]*[13579]\\..*",              devel: true }
 - { name: gnome-activity-journal,      verpat: ".*is.really.*",                            incorrect: true }
 - { name: gnome-applets,               verpat: "[0-9]+\\.[0-9]*[13579]\\..*",              devel: true }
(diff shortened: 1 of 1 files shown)


sternenseemann:
Is there a good way to check the substitutions on some example data? I've kind of had to write these patterns blind, so I can't quite be certain.

AMDmi3:
It's not correct co change existing official versions. You should add `verpat: "20[0-9]{2}", "sink: true"` rules instead so these are compared less than the actual versions. Also please squash into a single commit.

sternenseemann:
Okay, done! I've noticed two things we maybe should worry about:

* Some repositories carried the scheme forward when it was abandoned, e.g. Gentoo had a gnatcoll-core 2021 (but has abandoned this scheme for 22.0.0) and the AUR still uses it (2022 instead of 22.0.0).
* Debian has carried forward the scheme in a bizarre way where they publish 22.0.0 as 2022.0.0

AMDmi3:
> Some repositories carried the scheme forward when it was abandoned, e.g. Gentoo had a gnatcoll-core 2021 (but has abandoned this scheme for 22.0.0) and the AUR still uses it (2022 instead of 22.0.0).

These will be outdated as well but it's OK as we don't guarantee any sane handling for fake versions.

> Debian has carried forward the scheme in a bizarre way where they publish 22.0.0 as 2022.0.0

These can be ignored with additional rules.

> Is there a good way to check the substitutions on some example data? I've kind of had to write these patterns blind, so I can't quite be certain.

There are tests which are run in CI (you can run them locally with `make check`) which check the syntax of the file and RE's - this should be quite enough to not break anything. These don't check logic errors, bit it's not a big problem to correct these later. If you want to check against the real data you'll need to set up `repology-updater` - it's possible to use it in light mode to just parse and dump a few repos (`repology-update -ffpp <repos> && repology-dump <repos>` if you're curious), but it'll still require setting up the database, so IMO it's not practical for a one ruleset commit.

PR #655, opened 2022-11-14
Merge subprojects into gnatcoll-bindings, gnatcoll-db

sternenseemann:
The AdaCore repositories

* https://github.com/AdaCore/gnatcoll-bindings
* https://github.com/AdaCore/gnatcoll-db

contain a number of sublibraries (and executables) that are developed and released together. Some repositories split these up into their respective subprojects (nixpkgs, AUR) others package them all together (Fedora). For the former the projects gnatcoll-db and gnatcoll-bindings would not exist whereas it would be the only packages existing in the latter repositories.

To alleviate this, we just merge them all into those two names. As the libraries are interdependent it seems unlikely that mismatched versions can legitimately be in the same repository.

**Note** that this means for repositories like AUR and nixpkgs, multiple projects existing in the repository would be merged into one – I can't tell if that's a problem for repology.

(gnatcoll-db2ada and gnatinspect are currently named incorrectly in nixpkgs, but this will be fixed with
https://github.com/NixOS/nixpkgs/pull/201026/commits/ad3c52bf07a1c1295840fe3311cd53c390a6ab39)

### DIFF
--- a/800.renames-and-merges/g.yaml
+++ b/800.renames-and-merges/g.yaml
@@ -215,6 +215,21 @@
 - { setname: gn,                       name: generate-ninja }
 - { setname: gnash,                    name: gnash-git, ignore: true }
 - { setname: gnat-programming-studio,  name: [gnat-gps, gnatstudio] }
+- { setname: gnatcoll-bindings,        name: gnatcoll-gmp,      addflavor: gmp }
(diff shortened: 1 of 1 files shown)


AMDmi3:
Please add `addflavor: true` to these ([docs](https://github.com/repology/repology-rules/#addflavor)). We could also merge `libgnatcoll-bindings` and `libgnatcoll-db` (separate rules without addflavor).

PR #657, opened 2022-11-14
Mark 20YY.0.0 versions as incorrect for gprbuild

sternenseemann:
Correct would be YY.0.0, using the full "year" seems to be some repositories trying to carry an old scheme forwards incorrectly. This is only a problem for gprbuild at the moment, the gnatcoll-* packages are all imported correctly.


AMDmi3:
The pattern should've likely ended with `.*` to catch other patterns (like `2021.0.0.0778b109`), but I've thought it'd be easier to just extend sink pattern. Thank you!

PR #661, opened 2022-11-21
Disambiguate markdown ada module

Ikke:
A package from AUR called markdown, which is an ADA module, is conflated with the markdown application from daringfireball.


Ikke:
Alternatively we could add a wildcard rule for everything from (github.com)/AdaCore.

PR #667, opened 2022-12-08
Ignore atk 2.46.0 from gentoo

cemkeylan:
As [explained in Gentoo's repository](https://gitweb.gentoo.org/repo/gentoo.git/commit/dev-libs/atk?id=680a9a048f812387fd64f82d79cb533b1c9fedc1) ATK 2.46.0 published in Gentoo repositories is a dummy package for
at-spi2-core, and is incorrect. The latest version according to the [source repository](https://gitlab.gnome.org/GNOME/atk/-/tags) is 2.38.0


AMDmi3:
This won't have effect unless all versions are ignored (there's another one from termux). You should probably drop ruleset and ignore all versions greater than 2.38.0 (`vergt: "2.38" `).

PR #668, opened 2022-12-25
Add merge rules for pyenchant and its variants

tosky:


AMDmi3:
Incorrect naming should be fixed in repositories.

tosky:
Can you please explain? I was following what's in the README: "Rename or merge rules. Match a name, and set another name. The main purpose is to merge differently-named packages into the same project. For example, etracer, extremetuxracer, extreme-tuxracer → extreme-tuxracer."

Isn't this the same case? How does this differ from https://github.com/repology/repology-rules/pull/641/files or https://github.com/repology/repology-rules/pull/634/files ?

tosky:
If by "fixed in the repositories" you mean fixed in each distribution, and if that would really be possible, then repology wouldn't have its reason to exist.

AMDmi3:
The difference is that there are cases where naming discrepancies are tolerable and/or inevitable, such as where upstream is not completely clear regarding the project name spelling (etracer) or with variant packages (jitsi), and cases where name is clearly and unambiguously defined, so misnaming is not justified. Python modules fall to the latter category. We do not want to hide incorrectly named cases by merging them, we do not want to maintain rules for these cases, especially we don't want to introduce conflicts (such as with actual `enchant` module).

tosky:
This doesn't explain why other rules exist then, like the ones I pointed out and the many others which have been added even recently inside that same file. So what is the usefulness of repology then?

AMDmi3:
I don't see any point in explaining that.

PR #669, opened 2022-12-30
Add disambiguation for crane

Ikke:
Three distinct projects with the same name exist:

- https://github.com/michaelsauter/crane docker-compose like project
- https://github.com/ewilde/crane dotnet build tool
- https://github.com/google/go-containerregistry/tree/main/cmd/crane
  tool to interact with images and registries

### DIFF
--- a/850.split-ambiguities/c.yaml
+++ b/850.split-ambiguities/c.yaml
@@ -165,6 +165,11 @@
 - { name: cortex, wwwpart: cortexmetrics, setname: cortex-prometeus-storage }
 - { name: cortex, addflag: unclassified }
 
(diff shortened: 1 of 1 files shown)


AMDmi3 (changes requested):
- There are 3 projects
- There needs to be a final `unclassified` rule
- Might be better to reuse exiting name for google crane, probably `go:containerregistry`

Ikke:
@AMDmi3 Addressed your requested changes. Hope this is what you meant.

Ikke:
@AMDmi3 ping

AMDmi3:
Done, sorry for delay. Changed target names to more meaningful ones.

PR #679, opened 2023-02-14
Separate blackarch's snitch from others

tocic:
A BlackArch's tool called snitch is different from the modern https://github.com/cschreib/snitch. The latter is a testing framework. The former is a security-related program, and it was removed from the upstream. The removal from BlackArch may take a while, so it's worth adding an exception here.

See https://github.com/BlackArch/blackarch/issues/3708.

### DIFF
--- a/850.split-ambiguities/s.yaml
+++ b/850.split-ambiguities/s.yaml
@@ -257,6 +257,10 @@
 - { name: snappy, wwwpart: gnome, setname: snappy-player }
 - { name: snappy, addflag: unclassified }
 
+- { name: snitch, wwwpart: cschreib, setname: snitch-testing }
+- { name: snitch, wwwpart: ntsecurity, setname: snitch-security }
+- { name: snitch, addflag: unclassified }
+
 - { name: snoopy, wwwpart: tuxfamily, setname: snoopy-warcraft3 }
 - { name: snoopy, wwwpart: dssz, setname: snoopy-graph-anim }
 - { name: snoopy, wwwpart: [snoopylogger,a2o], setname: snoopylogger }


AMDmi3:
For 850* ruleset
- matching by ruleset is not allowed (so please match by wwwpat/wwwpart)
- rules for all counterparts and final catch-all `unclassified` rule are required

tocic:
I'd like to have the following behaviour: the original name `snitch` for every package except the BlackArch's one. IIUC, you suggest to use something like

```
- { name: snitch, wwwpart: ntsecurity.nu, setname: snitch-wine }
- { name: snitch, wwwpart: github.com/cschreib/snitch, setname: snitch }
- { name: snitch, addflag: unclassified }
```

But wouldn't that cause problems if the second package move to cschreib.github.io, for example? Or if another package with the same name but a different origin appears? Why can't we use "ignore" as a catch-all rule like here?

https://github.com/repology/repology-rules/blob/8fc78f15dac40462568ae6cb0c89c2830c801a61/850.split-ambiguities/s.yaml#L418-L420

Maybe just create a removal rool in 200.blacklists/* then? Perhaps in `upstreamless.yaml`.

```
- { name: snitch, ruleset: blackarch, remove: true }
```

AMDmi3:
> `- { name: snitch, wwwpart: github.com/cschreib/snitch, setname: snitch }`

It needs to be `setname: snitch-<something>` so `snitch` page shows disambiguation page instead of one of the projects, hiding other possible choices.

> But wouldn't that cause problems if the second package move to cschreib.github.io, for example?

That's why you should use just `cschreib` in the match.

> Or if another package with the same name but a different origin appears?

That is the whole point of this pattern - a new instance of known-to-be-ambiguous name gets into `foo-unclassified` and does not mess up existing projects.

The good question is what happens when another url appears for a known project. And the answer is that it also gets into `-unclassified`, which is lesser evil than messing up existing projects.

> Maybe just create a removal rool in 200.blacklists/* then?

No, removing a package would be providing incomplete project data. It's allowed only for non-project packages such as metapackages, and also for non-crossplatform software, only because I'm too lazy to do proper 850 rules for these.

tocic:
@AMDmi3 Thank you for the explanation, I've changed the rules.

PR #680, opened 2023-02-18
Ignore ver "0" and "deprecated" for vcpkg

dg0yt:
In vcpkg, use of version "0" indicates meta ports or ports which have been renamed.
Example: readline.

### DIFF
--- a/899.version-fixes.global.yaml
+++ b/899.version-fixes.global.yaml
@@ -68,6 +68,7 @@
 - { verpat: ".+20[0-9]{2}-[0-9]{2}-[0-9]{2}.*",                      ruleset: nix,         ignore: true }
 - { verpat: "[0-9a-f]{8}",                                           ruleset: pacstall, flag: git, rolling: true } # flag set in 100
 - { ver: "0",                                                        ruleset: vcpkg,       ignore: true } # deprecated, renamed, or meta port
+- { ver: "deprecated",                                               ruleset: vcpkg,       ignore: true }
 
 # nuget is terminally broken and can't handle arbitrary version schemes, so most are garbage
 - {                                                                  ruleset: chocolatey,  untrusted: true }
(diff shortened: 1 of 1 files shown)


dg0yt:
> Example: readline.

> These rules are useless and won't change any statuses. What are you trying to achieve?

I want version "0" of "readline" to be ignored for "vcpkg".
https://repology.org/projects/?search=readline&maintainer=&category=&inrepo=vcpkg&notinrepo=&repos=&families=&repos_newest=&families_newest=

AMDmi3:
This will not ignore readline. You've probably meant `remove` instead. However at least `dbhhelp` is neither of a deprecated, renamed, or meta port, so version criteria is unreliable. It may be better to check in the parser that portfile.cmake only contains `set(VCPKG_POLICY_EMPTY_PACKAGE enabled)` and comments or something.

dg0yt:
> However at least dbghelp is neither of a deprecated, renamed, or meta port, so version criteria is unreliable.

OTOH it doesn't really do much except for copying files out the Windows SDK. I could ask if they are willing to change to another value (date or just `1`).

> It may be better to check in the parser that portfile.cmake only contains `set(VCPKG_POLICY_EMPTY_PACKAGE enabled)` and comments or something.

This policy isn't really a good indicator. It can be used subject to `if()` guards, and it can be used even real ports such as `cmakerc` or  `coin-or-buildtools`.

AMDmi3:
> OTOH it doesn't really do much except for copying files out the Windows SDK. I could ask if they are willing to change to another value (date or just 1).

This won't change a thing and I don't endorse no-op changes just for repologys sake.

> This policy isn't really a good indicator. It can be used subject to if() guards, and it can be used even real ports such as cmakerc or coin-or-buildtools.

That's unfortunate. Well I suggest we don't do anything here then. After merging readline flavors, `readline` metapackage is considered `legacy` so it won't affect outdated packages lists and statistics. Other metapackages may be handled similar way (or remain unique packages, or be renamed in 850 rulese in case of clashes with unrelated projects). I don't like hiding stuff based on unreliable criteria.

dg0yt:
Thanks. Some things on the vpckg side are still a little bit immature. OTOH it is not always clear to me what is the desired resolution on the repology side, including where to draw the line between general pattern and specific rules.

PR #681, opened 2023-02-18
Handle vcpkg readline variants

dg0yt:


AMDmi3:
It doesn't look like osx and win32 variants are safe to merge.

PR #682, opened 2023-02-18
Map shiftmedia forks into origins

dg0yt:
These vcpkg packages are from forks with extra patches for building with MSVC. I expect them to lag, and there is no reason to hide it.


AMDmi3:
Reverted as these may have incompatible versions (see libgcrypt).

PR #687, opened 2023-04-03
Merge firefox versions

jopejoe1:
Merge a few Firefox versions into the main one.


jopejoe1:
Used package names instead of project names, updated it to use the project names now.

PR #690, opened 2023-04-13
Add thunderbird and fork merges

jopejoe1:
Add merges for thunderbird and its fork betterbird


AMDmi3 (changes requested):
- Variants (such as `thunderbird-bin`, `thunderbird-$lang`) need `addflavor` (as opposed of just different name spellings such as `mozilla-thunderbird`)
- Beta versions additionally need `weak_devel: true, nolegacy: true`

jopejoe1:
Updated to include `addflavor`, `weak_devel` and `nolegacy` to the place where needed.

PR #701, opened 2023-05-16
Add verignore for antic snapshots

jengelh:
MacPorts's antic is currently at "2022.11.30", which is not a tag found in the Git repository - those continue as normal with e.g. v0.2.5.


AMDmi3:
Please simplify regexp down to just e.g. `20[0-9]{2}.*`, and it should also be `incorrect` instead of `snapshot`. `snapshot` is for snapshots based on real versions

PR #702, opened 2023-05-16
Ignore additional Gentoo metapackage categories

mgorny:
Ignore the acct-* categories on Gentoo that are used to "install" user accounts, and app-alternatives that contains metapackages for switching between alternative implementations of software.

This should fix e.g. the incorrect Gentoo entry in:
https://repology.org/project/bc-unclassified

### DIFF
--- a/200.blacklists/metaports.yaml
+++ b/200.blacklists/metaports.yaml
@@ -4,6 +4,7 @@
 
 - { remove: true,   category: virtual, ruleset: [exherbo,gentoo] }
 - { remove: true,   category: java-virtuals, ruleset: [exherbo,gentoo] }
+- { remove: true,   category: app-alternatives, ruleset: gentoo }
 
 - { remove: true,   name: alias-font,   ruleset: freebsd }
 


200.blacklists/metaports.yaml
```diff
@@ -4,6 +4,9 @@
 
 - { remove: true,   category: virtual, ruleset: [exherbo,gentoo] }
 - { remove: true,   category: java-virtuals, ruleset: [exherbo,gentoo] }
+- { remove: true,   category: acct-group, ruleset: gentoo }
```
AMDmi3, line 7: You don't need `acct-*` here as these are ignored in the parsers

AMDmi3 (changes requested):

200.blacklists/metaports.yaml
mgorny: Ok, updated.

AMDmi3 (approved):

mgorny:
Thanks!

PR #716, opened 2023-06-19
README: Clarify split-ambiguities instructions

waldyrious:
Also link to the list of conditions that can be used to disambiguate packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ Here's a quick pointer of how to add specific rules:
 ### You want to split different projects with the same name
 
 - Open the corresponding yaml file under [`850.split-ambiguities/`](./850.split-ambiguities/)
-- Add a set of rules which distinguish packages, such as:
+- Add one or more rules which distinguish packages via specific [conditions](#conditions),
+  such as `wwwpart`, `category`, `verpat`, etc. For example:
   - `- { name: <ambiguous name>, wwwpart: <part of the homepage url>, setname: <specific name> }`
   - `- { name: <ambiguous name>, category: <category>, setname: <specific name> }`
   - `- { name: <ambiguous name>, verpat: <version pattern>, setname: <specific name> }`


AMDmi3:
I've updated these docs some more, please check.

waldyrious:
> I've updated these docs some more, please check.

I appreciate the extra detail, but it doesn't seem to match the actual content of the files. For example, in [850.split-ambiguities/f.yaml](https://github.com/repology/repology-rules/blob/master/850.split-ambiguities/f.yaml) which I just edited in #717, there are "groups" with a single entry, groups without a catch-all at the end, rules with conditions other than the three mentioned in the new docs (e.g. `ruleset`, `verpat`, `category`, etc.)

Perhaps the documentation is meant to describe the current best practices and the files still contain old/deprecated rule specifications? But even so, the discrepancy is likely to confuse people, so perhaps a FIXME comment (or something of the sort) in those lines would help.

AMDmi3:
Yes, existing groups without unclassified rule and `ruleset`/`verpat`/`category` rules are legacy, or some very rare special cases. I don't see any sense in adding a comment to each of these cases.

waldyrious:
Fair enough, but then how about changing "allowed" in

> wwwpart, wwwpat and sourceforge conditions are allowed

to "recommended" or "supported"? That way it doesn't contradict the presence of the legacy rules, which are deprecated but are technically allowed (i.e. the rule parser doesn't ignore or break on them). 

Alternatively, we could add a note like this:

> wwwpart, wwwpat and sourceforge conditions are allowed — other conditions may be found in the files but they are deprecated

WDYT?

AMDmi3:
It's a TL;DR section which is supposed to be as short as possible and describe exactly what should be done, without extra details. It currently lists exactly what should and may be done. People interested to explore ruleset may look into the rest of documentation for details.

waldyrious:
That's fine, but I can't seem to find that clarification (about supported/allowed conditions vs deprecated ones for the split-ambiguities ruleset) in the rest of the documentation either. Am I missing something?

Either way — without making this TL;DR more extensive, WDYT then about the first proposal of replacing "allowed" with "supported"? That way it's a bit less jarring to edit the file and find conditions that are not in the list.

AMDmi3:
> That's fine, but I can't seem to find that clarification (about supported/allowed conditions vs deprecated ones for the split-ambiguities ruleset) in the rest of the documentation either. Am I missing something?

You are not.

> Either way — without making this TL;DR more extensive, WDYT then about the first proposal of replacing "allowed" with "supported"? That way it's a bit less jarring to edit the file and find conditions that are not in the list.

Because it's not allowed, e.g. PR with such rules will not be accepted.

waldyrious:
Ok, got it. I disagree with the outcome but I can see why you'd prefer to have things this way. Thanks for your patience throughout this discussion.

PR #720, opened 2023-07-04
Cleanup pyside and shiboken

Bo98:
Currently we have:
```
pyside-tools
pyside2-tools
pyside6
pyside6-tools
python:pyside
python:pyside-2
python:pyside-2-tools
python:pyside-6
python:pyside-tools
python:pyside2
python:pyside2-tools
python:pyside6
python:pyside6-tools
```
Some are even miscategorised (WikiData reports 6.x under `python:pyside`).

This attempts to clean them up to look like (hopefully):
```
python:pyside
python:pyside-tools (not a flavor because different version scheme)
python:pyside2 (with tools as a flavor)
python:pyside6 (with tools as a flavor)
```

Similar situation applies to `shiboken` too.

I've not figured out setting up a local environment yet, so hopefully I've not made any mistakes.

### DIFF
--- a/800.renames-and-merges/python.yaml
+++ b/800.renames-and-merges/python.yaml
@@ -291,7 +291,9 @@
 - { setname: "python:pysha3",          name: pysha3 } # CVE
 - { setname: "python:pyside",          name: pyside }
 - { setname: "python:pyside",          name: [pyside-qt4,pyside-qt4-py3], addflavor: true }
-- { setname: "python:pyside2",         name: pyside2 }
+- { setname: "python:pyside-tools",    name: pyside-tools }
+- { setname: "python:pyside2",         namepat: "(?:python:)?pyside-?2(-tools)?", addflavor: $1 }
+- { setname: "python:pyside6",         namepat: "(?:python:)?pyside-?6(-tools)?", addflavor: $1 }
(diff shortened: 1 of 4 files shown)


800.renames-and-merges/python.yaml
```diff
@@ -291,7 +291,9 @@
 - { setname: "python:pysha3",          name: pysha3 } # CVE
 - { setname: "python:pyside",          name: pyside }
 - { setname: "python:pyside",          name: [pyside-qt4,pyside-qt4-py3], addflavor: true }
-- { setname: "python:pyside2",         name: pyside2 }
+- { setname: "python:pyside-tools",    name: pyside-tools }
+- { setname: "python:pyside2",         namepat: "(?:python:)?pyside-?2(-tools)?", addflavor: $1 }
+- { setname: "python:pyside6",         namepat: "(?:python:)?pyside-?6(-tools)?", addflavor: $1 }
```
Bo98, line 296: I could do:

```suggestion
- { setname: "python:pyside",         namepat: "(?:python:)?pyside-?[0-9]+(-tools)?", addflavor: $1 }
```

and then allow `950.split-branches.yml` to handle it so that future pysides are all in one place, if that makes sense? Particularly given that we need the 950 rules for WikiData, Homebrew etc anyway.

PR #721, opened 2023-07-05
Split dpp

Bo98:
Most referred to Discord bot library, but there was a dlang-cpp bridge tool there too.


AMDmi3:
Each split rule group should contain a rule for each target project, and a unclassified rule for not matched ones.

Bo98:
Does that mean a `setname` for everything too? Could probably do `dpp-discord` here.

PR #722, opened 2023-07-05
Split/merge distribution

Bo98:
A couple repos referred to one project and a couple referred to another: https://repology.org/project/distribution/versions

Split parts into `python:distribution` (which has had 3 different GitHub repo owners so far, so list them all).
Split others into `distribution-cncf`.
Merge in `docker-distribution` into `distribution-cncf`.

### DIFF
--- a/800.renames-and-merges/d.yaml
+++ b/800.renames-and-merges/d.yaml
@@ -92,6 +92,7 @@
 - { setname: diskonaut,                name: "rust:diskonaut" }
 - { setname: distcc,                   name: [distcc-gnome,distcc-gtk,distcc-monitor,distcc-nozeroconf,distcc-pump,distcc-server], addflavor: true }
 - { setname: distorm,                  name: [distorm3, distorm64, "python:distorm", "python:distorm3" ] }
+- { setname: distribution-cncf,        name: docker-distribution }
 - { setname: divide-and-succeed,       name: divideandsucceed }
 - { setname: djbdns,                   name: [djbdns-djb,djbdns-ipv4], addflavor: true }
 - { setname: djview,                   name: [djview,djview-qt4,djview-qt5,djview4,djvulibre-djview4], addflavor: true } # should be merged with djvulibre or not?
--- a/850.split-ambiguities/d.yaml
+++ b/850.split-ambiguities/d.yaml
@@ -126,6 +126,10 @@
 
 - { name: disktype, wwwpart: kamwoods, setname: disktype-kamwoods } # fork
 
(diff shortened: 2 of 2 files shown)


Bo98:
`-cncf` name picked from https://github.com/distribution

> CNCF Distribution Project

800.renames-and-merges/d.yaml
```diff
@@ -92,6 +92,7 @@
 - { setname: diskonaut,                name: "rust:diskonaut" }
 - { setname: distcc,                   name: [distcc-gnome,distcc-gtk,distcc-monitor,distcc-nozeroconf,distcc-pump,distcc-server], addflavor: true }
 - { setname: distorm,                  name: [distorm3, distorm64, "python:distorm", "python:distorm3" ] }
+- { setname: distribution-cncf,        name: docker-distribution }
```
AMDmi3, line 95: Please merge `docker-distribution` into `distribution` instead

It's more convenient and less error prone to make all packages follow the same ruleset path - that is everything is merged into a single name, then split. It requires less rules, and if we want to change the target name, it may be done in single place

850.split-ambiguities/d.yaml
```diff
@@ -126,6 +126,10 @@
 
 - { name: disktype, wwwpart: kamwoods, setname: disktype-kamwoods } # fork
 
+- { name: distribution, wwwpart: [philovivero, time-less-ness, wizzat], setname: "python:distribution" }
```
AMDmi3, line 129: Don't merge into `python:distribuition`

- That project may only contain corresponding pypi project (which is unrelated), use e.g. `distribution-tui-histograms`
- It's an application and not a python module, and these are usually merged into unprefixed name as it's how they are commonly packaged

850.split-ambiguities/d.yaml
```diff
@@ -126,6 +126,10 @@
...
+- { name: distribution, wwwpart: [docker, github.com/distribution/], setname: distribution-cncf }
```
AMDmi3, line 130: I'd pick a different target name, `cncf` is not descriptive. May reuse `docker-distribution`, or pick `distribution-docker` for consistency.

AMDmi3 (changes requested):

850.split-ambiguities/d.yaml
Bo98: Officially, it's no longer part of the Docker GitHub organisation and the "Docker" part of its name has been dropped: https://www.docker.com/blog/donating-docker-distribution-to-the-cncf/. It also not Docker-specific (e.g. GitHub Container Registry is supported).

Given it's an implementation for the OCI Distribution spec, would `distribution-oci` make sense?

PR #724, opened 2023-07-11
Mark shotcut 23.07.08 and 23.07.09 as devel

Ikke:
23.07.08, which has been renamed later to 23.07.09 are pre-release
versions, mark them as such.


AMDmi3:
Shouldn't it be 23.07.09 instead? I don't see 23.07.08 in the repository so it looks like `incorrect` rules should be added for it.

Ikke:
> Shouldn't it be 23.07.09 instead? I don't see 23.07.08 in the repository so it looks like `incorrect` rules should be added for it.

It appears they renamed the `23.07.08` release to `23.07.09`, it was correct (and existed) when I created this PR.

AMDmi3:
In that case you need rule for both versions as .08 is already used in the wild and .09 may appear in the future

Ikke:
> In that case you need rule for both versions as .08 is already used in the wild and packages for .09 may appear in the future

Done, I have marked them both as `devel` (I kept the `maintain` flag, I hope that's correct).

PR #726, opened 2023-07-22
rename gamedev-framework into gf-gamedev-framework

jube:
The package can also be named gf and the split rule now renames it gf-gamedev-framework instead of simply gamedev-framework, creating two project pages. This rename merge the gamedev-framework package into gf-gamedev-framework.


AMDmi3:
Your fix won't work as there's no `gf-gamedev-framework` at 800xx rule, because it's introduced in 850xx rule when `gf` is split. So instead, merge `gamedev-framework` into `gf` so all packages end up in the same project, and later split it into `gamedev-framework`, reusing the name used in vcpkg.

PR #754, opened 2023-11-06
renames & merges: add bitcoin-util

fanquake:
This binary is also shipped with Bitcoin Core. I assume `bitcoin-utils` is in here (and should remain) because some packagers are shipping the collections of "`bitcoin-cli`, `bitcoin-tx`, `bitcoin-wallet` & `bitcoin-util`" as `bitcoin-utils`.


AMDmi3:
There's no known project `bitcoin-util`, there's nothing to merge here.

PR #757, opened 2023-11-21
Merge buildah

FranklinYu:
https://packages.debian.org/sid/source/golang-github-containers-buildah

### DIFF
--- a/800.renames-and-merges/b.yaml
+++ b/800.renames-and-merges/b.yaml
@@ -207,6 +207,7 @@
 - { setname: bucardo,                  name: "perl:bucardo" }
 - { setname: bugs-everywhere,          name: bugseverywhere }
 - { setname: bugzilla,                 namepat: "bugzilla[0-9.-]+" }
+- { setname: buildah,                  name: "go:github-containers-buildah" }
 - { setname: buildbot,                 namepat: "buildbot[0-9.-]+" }
 - { setname: buildbot,                 name: "python:buildbot" }
 - { setname: buildbot,                 name: [buildbot-console-view,"python:buildbot-console-view"], addflavor: console-view }


AMDmi3:
Unfortunately we don't merge this naming pattern. go packages don't have canonical naming scheme so we'll have to manually maintain rules for each pattern of each package naming in order to merge them, which is not affordable. Additionally, it doesn't look like the url-based scheme is justified in this case.

FranklinYu:
Sorry I didn’t quite get the first part. Sounds like we have a general policy for (against?) Go packages? Is it documented anywhere, or is there prior discussions I can refer to?

AMDmi3:
https://repology.org/project/buildah/report

FranklinYu:
Thanks, I found the part against Go packages. But I still don’t quite get the reason: does naming-scheme refer to how the packager name the Debian package (i.e. `golang-github-containers-buildah`)? I thought that packages are merged one by one anyway, even with Python or Perl. How would the lack of consistent naming scheme make the maintenance more difficult?

FranklinYu:
You mean, for other languages (Python and Perl) the wildcard rules can handle most packages, while this pattern cannot scale with Go modules? This is the concern of accepting Go merges?

Then maybe in this specific case (Buildah) there is another way to fix it: the Debian *source* package is the bad name [`golang-github-containers-buildah`](https://packages.debian.org/source/sid/golang-github-containers-buildah), but one of the *binary* package is simply [`buildah`](https://packages.debian.org/sid/buildah). Would that help us decreasing the number of manual merges required for some “(relatively) good” Go packages?

AMDmi3:
> You mean, for other languages (Python and Perl) the wildcard rules can handle most packages, while this pattern cannot scale with Go modules? This is the concern of accepting Go merges?

Yes.

> Then maybe in this specific case (Buildah) there is another way to fix

Not really. Binary packages have their own name discrepancies and require more manual rules on their own.

PR #760, opened 2023-12-15
rimgo: marked date verpat as incorrect

quantenzitrone:
Should fix the `2022.04.22` version from the AUR being marked as newer than `1.2.1`.


AMDmi3:
That should be `incorrect` instead. `snapshot` is for versions based on real upstream releases.

PR #769, opened 2023-12-30
Merge libsass-python into python-libsass

FabioLolix:
https://repology.org/projects/?search=libsass

https://repology.org/project/libsass-python/versions

https://github.com/sass/libsass-python

Hello, Debian and derivates are using upstream repository name while all the other are using `python-libsass`


AMDmi3:
We don't merge python modules not clearly prefixed as such, sorry.

PR #770, opened 2023-12-30
Go wildcard for GNU Guix

FabioLolix:
Hello, not sure if this is correct

With a generic search for `libsass`  https://repology.org/projects/?search=libsass

There are `go:github-bep-golibsass` for Fedora and DeBuntu and `go:github-com-bep-golibsass` for GNU Guix

* https://repology.org/project/go:github-bep-golibsass/versions
* https://repology.org/project/go:github-com-bep-golibsass/versions

Upstream is the same https://github.com/bep/golibsass https://github.com/bep/golibsass/tags

Searching for `github-com` https://repology.org/projects/..go:github-com-biogo-store-step/?search=github-com

There are many results for not normalized Go modules from GNU Guix


AMDmi3:
Rules for Go modules will not be accepted, as there's no canonical naming scheme and there's no possibility to convert between schemes.

PR #776, opened 2024-01-21
libdigidocpp, qdigidoc4:  Ignore Arch version

klemensn:
It is the only system that adds a fourth version component to the official x.y.z schema, thus falsely marking other systems as outdated, e.g. Arch libdigidocpp-3.16.0.1442 vs. OpenBSD libdigidocpp-3.16.0.


klemensn:
Not sure what's appropiate here (`ignore` or `incorrect`?) and how that'd effect the listing.
Arch's package should still be listed and be comparable, but not cause others to seem outdated.

I maintain these packages on OpenBSD seeing them under https://repology.org/projects/?maintainer=kn@openbsd.org&inrepo=openbsd&outdated=1 despite matching upstream's latest releases is annoying.

klemensn:
FYI @kevku presuming you're the same `kevku` mentioned as Arch's package maintainer: 
https://aur.archlinux.org/packages/libdigidocpp
https://aur.archlinux.org/packages/qdigidoc4

AMDmi3 (changes requested):

PR #778, opened 2024-01-26
merge libtg-owt into tg-owt

klemensn:


klemensn:
The majority of versions use the commit date, e.g. `20210320`, some as version component, with a prefix.
Would it make sense to munge those prefixes/suffixes, drop `noscheme = true` and use the date as de-facto version?

https://repology.org/project/tg-owt/versions

AMDmi3:
Thanks. I've also merged some more instances.

> Would it make sense to munge those prefixes/suffixes, drop noscheme = true and use the date as de-facto version?

We don't do that. Upstream must provide official versioning scheme.

PR #782, opened 2024-02-17
Disambiguate noson vs. noson-app

klemensn:
https://github.com/janbar/noson is the library used by
https://github.com/janbar/noson-app Qt5 GUI program.

All packages on repology stick this naming scheme except for nixpkgs
calling the GUI just noson, thus falsely marking all other noson library
packages out of date (noson-app 5.6.5 vs. noson 2.12.6).


AMDmi3:
Thanks, but I'd rather add an ignore rule here, as it looks more like a misnamed package.

PR #784, opened 2024-02-28
Add version ignore and untrusted for tinyxml

mhoush:
The wikidata version for "tinyxml" is actually "tinyxml2". I see related comments about this in https://repology.org/project/tinyxml/report, here's a pull request.

Based on https://github.com/repology/repology-rules?tab=readme-ov-file#rule-syntax I think this is the proper fix but please advise if it's not.

Thanks!

### DIFF
--- a/900.version-fixes/t.yaml
+++ b/900.version-fixes/t.yaml
@@ -139,6 +139,8 @@
 - { name: tinyxml,                     verpat: "2-.*",                                     incorrect: true } # nix
 - { name: tinyxml,                     ver: "2.6.2_2",               ruleset: buildroot,   incorrect: true }
 - { name: tinyxml,                                                   ruleset: buildroot,   untrusted: true } # accused of fake 2.6.2_2
+- { name: tinyxml,                     ver: "10.0.0",                                      incorrect: true } # tinyxml2
+- { name: tinyxml,                                                   ruleset: wikidata,    untrusted: true } # wikidata reports version from tinyxml2 instead
 - { name: tivodecode,                  ver: "0.2",                   ruleset: fedora,      incorrect: true }
 - { name: tivodecode,                                                ruleset: fedora,      untrusted: true }
 - { name: tk,                          verpat: ".*a[0-9]+",                                devel: true }


AMDmi3:
I've seen the reports on this but still haven't decided.
Option a is to merge tinyxml with tinyxml2 because the former is deprecated and the latter is successor.
Option b is to fix wikidata.
Adding ignore for wikidata is not an option.

Are there any objections against option a?
(also closing all reports and redirecting here)

mhoush:
It does seem that the former is deprecated but at the same time, tinyxml2 is not backwards-compatible with tinyxml, it seems. At least kodi for example will not build against tinyxml2.

mhoush:
To clarify my thinking with the ignore rule, my understanding is that this proposed change would only ignore wikidata information for the "tinyxml" project but NOT for "tinyxml2".

Was that the wrong way to think about it?

AMDmi3:
> It does seem that the former is deprecated but at the same time, tinyxml2 is not backwards-compatible with tinyxml, it seems. At least kodi for example will not build against tinyxml2.

Yes, but in repology it can be processed as any other library with multiple [possibly also incompatible] versions, where multiple versions are allowed to coexist, older is marked as legacy as long as its latest in its branch (1.x that is) and later version (2.x, that is) is also provided.

> To clarify my thinking with the ignore rule, my understanding is that this proposed change would only ignore wikidata information for the "tinyxml" project but NOT for "tinyxml2".

That is correct. But ignoring freely editable source instead of fixing it is incorrect at its very root.

mhoush:
Fair enough... so the preferred course would be both options a and b, I guess.

AMDmi3:
I've already told this multiple times. It does not have priority and is not used as a reference. It's that a) a problem should be fixed at its root (not ignored and especially not spent dedicated effort on to be handled downstream) b) anyone can do it right away (not wasting my time for doing that for them and wasting their time for waiting it to be fixed).

jrmarino:
>  It does not have priority and is not used as a reference.

The evidence is against this.  If the wiki is the ONLY reference that says tinyxml is at version 10.0.0, and that all repositories are out of date because 10.0.0 is considered the latest, then of course the wiki is used as a reference.

> a problem should be fixed at its root 

This should go without saying, but I don't want to be told to edit the wiki.  It's not my job.  The wiki is shown to be wrong FREQUENTLY.  It sucks.  (frankly I don't even know how to get information out of it).

So rather that "fix it at the root" which is nobody's job except the people that host the wiki, it should be marked untrusted because it's not trustworthy.  Frankly I'd rather it be dropped completely.

It's your choice to include this as a source, so it's your time that should be spent when that choice causes a problem.

don't bother to respond.  I already know you disagree.

AMDmi3:
In the meantime, I've merged tinyxml and tinyxml2 entries, which fixed the issue with tinyxml being outdated by tinyxml2 versions.

PR #785, opened 2024-02-28
c.yaml: merge cargo-auditable package names into one

Shnatsel:
This merges various names of my project, https://github.com/rust-secure-code/cargo-auditable

The names currently in use can be found [here](https://repology.org/projects/?search=cargo-auditable) so you can double-check my edit.

Thanks!


AMDmi3:
There is no `rust-cargo-auditable`  project known, you may only leave `rust:cargo-auditable`. You may also shorten it to `rust:$0` and drop quotes

PR #786, opened 2024-03-01
Merge ldid (from ProcursusTeam)

TheRealKeto:
[The Procursus Team](https://github.com/ProcursusTeam) maintains a fork of an original project named ldid as `ldid-procursus`. In Repology, some entries within the `ldid` project aim to the version scheme used by the Procursus Team (causing them to appear to be outdated). 

Instead, these entries (notably from the AUR), now or in the future (from other package managers), are to be considered part of the `ldid-procursus` project.

### DIFF
--- a/800.renames-and-merges/l.yaml
+++ b/800.renames-and-merges/l.yaml
@@ -39,6 +39,7 @@
 - { setname: lcms,                     name: [lcms,lcms1,lcms2,little-cms,little-cms2,lcms-python,liblcms,liblcms2,littlecms] }
 - { setname: ldb,                      namepat: "(?:lib)?ldb[0-9]*" }
 - { setname: ldc,                      name: liblphobos, addflavor: true }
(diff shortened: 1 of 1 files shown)


AMDmi3:
Is the original project still maintained? At least its repo looks dead.

The change is not correct, as such rule belongs to 850 ruleset and requires counterpart to match the original project and a fallback. However it looks like in this case it would make more sense to instead merge into a single project and make procursus versions compare higher than original.

TheRealKeto:
Thanks for the feedback!

In regards to the original project vs the one maintained by the Procursus Team, a rationale behind the projects existing with different names comes from [a comment on a pull request submitted to replace the original project with the aforementioned fork on Homebrew](https://github.com/Homebrew/homebrew-core/pull/136859#issuecomment-1642590158).

In other words, the original project (`ldid`) is still maintained, rather than abandoned, so the best case is to keep the packages separate. Unsure if this is applicable to Repology, so the approach is entirely up to you.

PR #791, opened 2024-03-20
Disambiguate `dawg` on Windows from Debian package

Trenly:


AMDmi3:
This needs split rules for all variants and also catch-all `unclassified` rule.

Trenly:
> This needs split rules for all variants and also catch-all `unclassified` rule.

I'm new to this repo, so I don’t know what most of that terminology means. Is there any better documentation or examples? The README wasn’t entirely clear on what needs to be added for each rule, just the different fields available

AMDmi3:
See around in the 850 ruleset - there are still a few exceptions which doesn't follow the pattern, but it should be clear nonetheless.

There should be a split rule which gives a new unique name for each individual project with a given ambiguous name, and there should be rule which sets `unclassified` flag on packages which haven't matched any of these rules. So you need a rule for `wwwpat: reedacartwright` and `addflag: unclassified` rule. Note that it's not necessary to add rules for dead homepages, such as `scit.us/projects/dawg` in this case.

Trenly:
> See around in the 850 ruleset - there are still a few exceptions which doesn't follow the pattern, but it should be clear nonetheless.
> 
> There should be a split rule which gives a new unique name for each individual project with a given ambiguous name, and there should be rule which sets `unclassified` flag on packages which haven't matched any of these rules. So you need a rule for `wwwpat: reedacartwright` and `addflag: unclassified` rule. Note that it's not necessary to add rules for dead homepages, such as `scit.us/projects/dawg` in this case.

Thank you for the guidance, I appreciate it

AMDmi3:
Sorry, forgot to mention - target names are usually prefixed with original name for them to remain grouped and in the same location of project list. That is, `dawg-digital-audio-workstation-group` and `dawg-dna-assembly-with-gaps` or with shorter suffixes. Otherwise good.

PR #793, opened 2024-03-22
Fix Anope 2.1 not being marked as a development release.

SadieCat:
Anope has a stable branch (2.0) and a development branch (2.1). Fedora (and its descendants) decided to ship the development branch for some reason. This has resulted in Repology marking distributions that ship the stable version as out of date.


AMDmi3:
Is it only 2.1 or all odd minor versions?

SadieCat:
Its all odd minor versions. Is there a better way to express that?

AMDmi3:
Yes, see [13579] pattern for KDE projects (900/k.yml)

SadieCat:
I've updated this PR to use the same matching logic.

PR #795, opened 2024-03-31
Add librustls rules

kpcyrd:


AMDmi3:
Is there reason to chose `librustls` name? Both upstream project name and crate name are `rustls-ffi`. `rustls-ffi` is more widely used package name too.

kpcyrd:
Updated!

PR #797, opened 2024-04-02
Merge swiftlang

jbicha:
Debian has packaged swift-lang using the source package name [swiftlang](https://tracker.debian.org/pkg/swiftlang).

This is untested and I'm unsure if it's the right fix. Note that 850.split-amgibuities/s.yaml renames swift to swift-lang if there is a URL containing swift.org which the Debian package should have. So it's unclear whether it is better to just set the correct name (swift-name) directly and on the previous line for swift-language or set it to swift and let the ambiguity rule handle it afterwards.


jbicha:
Done. Thanks

PR #807, opened 2024-04-19
Merge libproxy subpackages into libproxy

DimStar77:


AMDmi3:
> I hope I got the format right

Yes, but there's no need for a separate file for it. Please do a single line rule in 800/lib

PR #809, opened 2024-04-21
mumble: 1.5.x up to 1.5.629 are RC

andreasstieger:
mumble: 1.5.x up to 1.5.629 are RC
See https://www.mumble.info/blog/mumble-1.5.629-rc-3/
https://www.mumble.info/blog/new-versioning-scheme/


AMDmi3:
It makes sense to use `verge`/`verle` or `rel`* instead of pattern here. Also please add `maintenance: true` to indicate that this needs manual maintenance.

PR #811, opened 2024-04-30
Rename emulationstation-de to es-de

bentley:
According to the upstream FAQ
(https://gitlab.com/es-de/emulationstation-de/-/blob/master/FAQ.md):

“As of the 3.0.0 release the official name for the project and application is ES-DE Frontend or simply ES-DE.”


bentley:
This is really a merge, not just a rename, because Repology has the same project under multiple names:

- https://repology.org/project/es-de/versions
- https://repology.org/project/emulationstation-de/versions

PR #815, opened 2024-05-16
Add verignore for fte

Sigmanificient:
The version of fte is not `20110708` but `0.50.2`, however there is a confusion which the sourceforge archive that has the upload date. The version can be found within the common archive, inside the `NEWS` file:

```
$ cat NEWS
Revision history
-----------------------------------------------------------------------
0.50.2b6 -- <fill month here> 2011
  + Allow \r at the end-of-line when looking for the end-of-heredoc in
    Perl hilighting mode to make it easier to read files created on windows
    and transfered to unix incorrectly.
  + Add ShowTildeFilesInDirList global config option to allow hiding
    of *~ in the directory list (most of the time, I don't want this)
    default is true.
  ! Allow open-brace on same line as function name in shell mode
  + Added permission flags to the directory listings on unix to show rwx
    at a glance - makes it easier to debug certain issues, know if you
    can write to the file, etc.
  ! Compilation fixes for Windows compilers
  + Allow JAVA colorizer to be used for javascript files (.js)
  - Ongoing conversion of return codes to return consistently
    1 for success or 0 for failure
  ! Colorizing man pages again (using MAN_KEEP_FORMATTING)
  + adding 2nd FD for WaitFdPipeEvent() so used in con_linux.cpp

0.50.2b5 -- March 2010
  + Allowed -m flag to take lower-cased version of mode names
  ! .t files are perl test files, added.
...
```

### DIFF
--- a/900.version-fixes/f.yaml
+++ b/900.version-fixes/f.yaml
@@ -324,6 +324,7 @@
 - { name: fstrcmp,                     verpat: "[0-9]+\\.[0-9]+\\.D[0-9]+",                any_is_patch: true }
 - { name: fstrcmp,                     verpat: "[0-9]+\\.[0-9]+\\.[0-9]+",                 incorrect: true }
 - { name: fswebcam,                    vergt: "20140113",            ruleset: sisyphus,    snapshot: true } # https://www.sanslogic.co.uk/fswebcam/
+- { name: fte,                         verpat: "[0-9]{8}",                                 altscheme: true } # sourceforge publish date
 - { name: ftgl,                        ver: "2.1.3",                 ruleset: [gobolinux,mageia,openmandriva,pclinuxos,pld], incorrect: true } # it's 2.1.3-rc5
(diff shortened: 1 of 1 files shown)


AMDmi3:
According to this changelog, 0.50.2 is not valid either.

Sigmanificient:
it is `0.50.2b6`?

AMDmi3:
I think it should remain 20110708 as least ambiguous.

Sigmanificient:
Most repository seems to follow the `0.50.2` approach, even `0.50.2b6-20110708`: https://repology.org/project/fte/versions
However they are all marked as outdated due to the 2 repositories that used `20110708`

AMDmi3:
In fact that's 2x`20110708` vs 2x`0.50.2` vs 1x`0.50.2b6` as counters on the page say because for instance all debian derivs are a single family. `0.50.2b6-20110708` is made up.

We can mark dates as altscheme so we'll have both `20110708` and `0.50.2` as latest.

Sigmanificient:
> We can mark dates as altscheme so we'll have both `20110708` and `0.50.2` as latest.

That would be nice!

AMDmi3:
There's problem in rule syntax.

PR #816, opened 2024-05-21
Use common name for Sherlock Project

ppfeister:
The common and recognized name for Sherlock is either Sherlock or Sherlock project.
This PR should just fix the prior ambiguity split's name of sherlock-social-osint.

### DIFF
--- a/850.split-ambiguities/s.yaml
+++ b/850.split-ambiguities/s.yaml
@@ -152,7 +152,7 @@
 - { name: shed, addflag: unclassified }
 
 - { name: sherlock, wwwpart: open-music-kontrollers, setname: lv2:sherlock }
-- { name: sherlock, wwwpart: sherlock-project, setname: sherlock-social-osint }
+- { name: sherlock, wwwpart: sherlock-project, setname: sherlock-project }
 - { name: sherlock, wwwpart: sergius02, setname: sherlock-ip-info }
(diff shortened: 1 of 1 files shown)


AMDmi3:
`project` suffix for a project is a tautology and does not help to resolve ambiguity.

ppfeister:
Hey Dmitry

That was a quick reply

I agree with you, but not necessarily in this situation. If `project` was an arbitrary suffix, then yes, it's absolutely pointless. However, when it's part of the actual name, it increases specificity and actually helps.

As an example, the foundation called Free Software Foundation doesn't lose the last word of the name, becoming simply Free Software, due to the word's use as a description elsewhere. It's a proper noun rather than an adjective in this case.

ppfeister:
As an alternative...

`sherlock-project` is already in use on a handful of packages due to name parity requirements and such (i.e. Fedora's required parity with PyPI). It stems from the name of the project's linked org. Another name was considered, however, for when `sherlock` was unavailable - `sherlock-search`.

This has yet to be used anywhere else, though, which makes it pretty equally poor, but it is an option that exists

AMDmi3:
These suffixes are used for disambiguation within repology and have nothing to do with upstream alternative naming choices. `project` is useless and `search` is still less specific than the current one.

PR #826, opened 2024-06-30
Split satellite

Luflosi:
I'm not 100% sure if I used `wwwpart` correctly. Is there some way to test the rules locally?

Do the new names seem reasonable?


Luflosi:
Done.

PR #827, opened 2024-07-10
Update Sherlock Proj

ppfeister:
Sherlock now uses sherlockproject[.]xyz rather than sherlock-project[.]github[.]io in some locations. This change will merge any packages that stray by using the new domain. Currently, one seems to have separated (Homebrew).

This also merges the Fedora package, sherlock-project, with the rest of the relevant packages at sherlock-social-osint, reducing 3 apparent projects to 1.

___

__Aside:__

While opening this PR, I had wondered... is there anything actually making this project use sherlock-social-osint rather than just sherlock? There is a python:sherlock, but this isn't in any real conflict as one is a library, therefore prefaced with python:, and the other is an application, almost always without preface when presented.

https://github.com/repology/repology-rules/blob/6192111764810e93150afd5bbd6bcc9ecf9dea5f/800.renames-and-merges/s.yaml#L147

This snippet from two months ago seems to agree with reverting the name to simply `sherlock`

(figured I'd ask while PRing anyways)

### DIFF
--- a/800.renames-and-merges/s.yaml
+++ b/800.renames-and-merges/s.yaml
@@ -145,6 +145,7 @@
 - { setname: shellcheck,               name: shellcheck-bin, addflavor: true }
 - { setname: shellcheck,               name: "haskell:shellcheck" }
 - { setname: sherlock,                 name: python:sherlock-project }
+- { setname: sherlock-social-osint,    name: sherlock-project }
 - { setname: shfmt,                    name: $0-bin }
 - { setname: shibboleth-sp,            namepat: "shibboleth-sp[0-9]+" }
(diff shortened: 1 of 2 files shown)


AMDmi3:
Please merge into `sherlock` so it comes through the split rule.

> is there anything actually making this project use sherlock-social-osint rather than just sherlock?

Yes,`sherlock` is split into at least [two](https://repology.org/project/sherlock/versions) more projects.

ppfeister:
Ah, I see what you're doing now. Didn't quite register the first time I read it, seeing just 'sherlock'.

Updating now.

PR #828, opened 2024-07-10
Fix Sherlock pattern

ppfeister:
Fixes pattern for fullmatch.


ppfeister:
My bad -- didn't realize it was using fullmatch rather than search. Wrongly assumed it was sub match acceptable like with wwwpart.

AMDmi3:
Let's convert this into wwwpart array, it would be more straightforward.

ppfeister:
That should work. Simpler that way.

PR #831, opened 2024-07-20
Fix senpai-irc classification for new upstream location

supertassu:
~~The chess engine is no longer packaged in AUR (nor was it ever packaged
anywhere else), so this is no longer necessary. The classification was
already a bit broken due to the upstream move from
[git.sr.ht/~taiite/senpai](https://git.sr.ht/~taiite/senpai) to [git.sr.ht/~delthas/senpai](https://git.sr.ht/~delthas/senpai/),
but I think removing it entirely is a better option than fixing it.~~

The upstream repository recently moved from https://git.sr.ht/~taiite/senpai to https://git.sr.ht/~delthas/senpai.

### DIFF
--- a/850.split-ambiguities/s.yaml
+++ b/850.split-ambiguities/s.yaml
@@ -119,7 +119,7 @@
 - { name: selene, addflag: unclassified }
 
 - { name: senpai, wwwpart: chess, setname: senpai-chess }
-- { name: senpai, wwwpart: taiite, setname: senpai-irc }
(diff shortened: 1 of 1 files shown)


AMDmi3:
Let's fix it instead. Split rules should never be removed, as the name is known to be ambiguous, and the chess engine package may reappear anytime.

supertassu:
Sure, updated.

PR #833, opened 2024-08-01
Merge readability with python:readability-lxml

City-busz:
Actually it's one project:
https://pypi.org/project/readability-lxml/
https://github.com/buriy/python-readability


City-busz:
But it's obviously a python module, and the binary package is called [python3-readability](https://packages.debian.org/sid/python3-readability) on Debian.

AMDmi3:
Source package (and repology uses source packages) is called `readability`, and that's exactly why it is `readability` and not `python:readability`. Btw, now I see that as it should be `python:readability-lxml`, it's also incorrectly named.

PR #835, opened 2024-08-07
Handle somen nixpkgs package sets

jopejoe1:
Sorted through a few packages sets in nix.


AMDmi3:
Sorry, adding wildcard rules for ecosystems not packaged anywhere else is pointless, and I do not plan to extend nix categories whitelist - there should be a way to know whether a package belongs to a module ecosystem, with that there will be no need for the whitelist at all.

LunNova:
> Sorry, adding wildcard rules for ecosystems not packaged anywhere else is pointless, and I do not plan to extend nix categories whitelist - there should be a way to know whether a package belongs to a module ecosystem, with that there will be no need for the whitelist at all.

Is there anything actionable here on the nixpkgs side / more specifics on how our metadata needs to change to help repology, or is "there should be a way to know whether a package belongs to a module ecosystem" something that needs handled on the repology side?

AMDmi3:
Let me revisit the problem first:

It's common for nix to add new categories of packages, which can be vaguely described as modules for some programming languages (like `python313Modules`, `perl5Modules`) or plugins/extensions for some software (like `kodiPackages`). It was not guaranteed for these packages to have distinctive name prefix, and after ongoing switch to `pname` there will be no prefixes available at all, and without prefix or some special handling these packages are very likely to cause conflicts with other software (for instance, php zstandard module ending up in a same project as zstandard C library).

So to avoid such conflicts we need to reliably identify current and future categories of packages which require separate namespace.

For this purpose, we currently take the part of the package key (IIRC it's called "derivation name"?) up to the first period, call that a `category`, and use it in the ruleset to add namespace prefix to the projects (such as python: for `python313Packages`). That works relatively well, but we also need to prevent packages from newly added categories for which we don't have rules yet from ending up without namespace and causing conflicts, so we pessimize these to `-unclassified` projects, which is intended to induce reports which would lead to adding new namespaces.

The problem, however, is that not all categories need namespaces. For instance, where's `llvmPackages` or `kdePackages` which do not cause conflicts and do not need any namespace. There's also `pidginPackages`, which is a plugin collection from one hand, bit it also contains pidgin itself and all plugin pnames are prefixed with `pidgin-` so it also doesn't need any namespace.

Some heuristics are used to decide whether to pessimize a category:
https://github.com/repology/repology-rules/blob/f4d2bbc22ad50a45943541ea094327776fbc82ec/899.version-fixes.global.yaml#L56-L67
but these require keeping a whitelist of categories which don't need a namespace, which is not maintainable.

So, an actionable item on nix side would be to provide some hint on whether a package needs namespace. I believe it's already solved on nix side which should probably have the same aliasing problem within itself, and manifests itself as `name` (as opposed to `pname`) prefixes. I'm pretty sure though that I've had multiple cases of conflicts on unprefixed package names before I had to introduce the scheme described above, so I need to recheck it.

Ideally, presence of `name` prefixes would correspond to what I need, and exposing them explicitly would be sufficient. If there are exceptions, we'll need to think of how to handle these.

PS. I should reconsider this PR. Expanding a category whitelist is still not desirable, but adding new wildcard rules now looks like the right solution regardless of whether the projects in a named category are packaged anywhere else.

PR #839, opened 2024-08-17
Merge mwclient into python:mwclient

supertassu:


AMDmi3:
Sorry, we don't merge module packages which lack naming scheme allowing to identify them as such. Unfortunately that's common problem for Debian.

PR #843, opened 2024-09-08
Remove outdated verignore for ccls

cho-m:
0.20240202 exists - https://github.com/MaskRay/ccls/releases/tag/0.20240202

I think this rule is incorrectly marking majority of repositories as on an "incorrect version" https://repology.org/project/ccls/information


AMDmi3:
For the record, the release was created after (Apr 19) the report and the rule (Mar 22). Other verignores for this project  are still valid.

PR #844, opened 2024-09-09
Add sgt-puzzles rules for Alpine and Chimera

EarldridgeJazzedPineda:


AMDmi3:
We don't add rules for incorrect versions, these should be fixed in repositories.

PR #847, opened 2024-09-11
Add merge rules for Chimera Linux

EarldridgeJazzedPineda:


800.renames-and-merges/e.yaml
```diff
@@ -78,6 +78,7 @@
 - { setname: emacs,                    name: [emacs-devel-nox], addflavor: nox11, weak_devel: true, nolegacy: true }
 - { setname: emacs,                    name: emacs-git-gtk2, addflavor: gtk2, ignore: true }
 - { setname: emacs,                    namepat: "emacs[0-9.-]+(-non-dfsg)?" }
+- { setname: emacs,                    name: emacs-console }
```
AMDmi3, line 81: addflavor

AMDmi3 (changes requested):

800.renames-and-merges/e.yaml
EarldridgeJazzedPineda: There isn't a package called `emacs` in Chimera Linux, and emacs is called `emacs-console` in that repository. `addflavor` isn't applicable here.

800.renames-and-merges/linux.yaml
```diff
@@ -38,6 +38,7 @@
     - linux-headers
     - linux-mainline
     - linux-tools
+    - linux-stable
```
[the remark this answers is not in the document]
EarldridgeJazzedPineda: There are multiple Chimera Linux packages that belong to Repology's `linux` package, so I added it into the `addflavor` group just to be safe.

https://repology.org/project/linux/versions

800.renames-and-merges/e.yaml
AMDmi3: Ruleset does not need to be aware of whether there are other flavors of the same project in repository, and also this can change in the future. So if a name looks flavored, it needs `addflavor`.

800.renames-and-merges/linux.yaml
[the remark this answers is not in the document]
AMDmi3: Packages varying solely on version are not flavors.

AMDmi3 (approved):

PR #849, opened 2024-09-11
Mark python:kaleido 0.2.1.post1 as setver: "0.2.1"

Pandapip1:


AMDmi3:
It's not incorrect as its official vesion. You can instead use `setver: "0.2.1"`.

Pandapip1:
Fixed!

PR #850, opened 2024-09-11
Disambiguate envision

Pandapip1:
Turns out the envision package was actually four

### DIFF
--- a/850.split-ambiguities/e.yaml
+++ b/850.split-ambiguities/e.yaml
@@ -97,6 +97,12 @@
 
 - { name: enigma, wwwpart: sgtatham, setname: enigma-sgtatham }
 
+- { name: envision, wwwpart: gabmus, setname: envision-xr }
+- { name: envision, wwwpart: atari.miribilist.com, setname: envision-atari }
+- { name: envision, wwwpart: enlightenment.org, setname: envision-video-player }
+- { name: envision, wwwpart: ethz.ch, setname: envision-ide } # AUR package doesn't have a homepage
+- { name: envision, addflag: unclassified }
+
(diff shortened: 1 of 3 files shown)


Pandapip1:
Does it want it in alphabetical order?

850.split-ambiguities/e.yaml
```diff
@@ -97,6 +97,12 @@
 
 - { name: enigma, wwwpart: sgtatham, setname: enigma-sgtatham }
 
+- { name: envision, wwwpart: gabmus, setname: envision-xr }
+- { name: envision, wwwpart: atari.miribilist.com, setname: envision-atari }
+- { name: envision, wwwpart: enlightenment.org, setname: envision-video-player }
+- { name: envision, ruleset: aur, setname: envision-ide } # AUR package doesn't have a homepage
```
AMDmi3, line 103: ruleset matchers are not allowed here. You can use wwpart: `dimitar-asenov` or `ethz.ch` as there are related homepages https://github.com/dimitar-asenov/Envision.

850.split-ambiguities/e.yaml
```diff
@@ -97,6 +97,12 @@
 
 - { name: enigma, wwwpart: sgtatham, setname: enigma-sgtatham }
 
+- { name: envision, wwwpart: gabmus, setname: envision-xr }
```
AMDmi3, line 100: Please also add `envision-xr` -> `envision` rule in 800/x ruleset, so all packages come through this rule, and are affected if it's changed.

AMDmi3 (changes requested):

PR #851, opened 2024-09-11
Merge immersed into immersed-vr

Pandapip1:


AMDmi3:
I don't see anything which would justify -vr suffix, so I suggest to fix the name in nix instead.

PR #852, opened 2024-09-11
Merge paup into paup-cli

Pandapip1:


AMDmi3:
Since there are cli and gui variants, `paup` name is ambiguous, the rule should be reversed and placed in 900 ruleset instead. And since neither repository provides download url which could be used to split it, I guess that there's nothing we can do about this project.

PR #855, opened 2024-09-16
Merge gnome-sushi with sushi

City-busz:
gnome-sushi is a Debian-specific name of this app.


AMDmi3:
`sushi` name is ambiguous and clashes with another project, that's why debian uses different name. You should submit split 850 ruleset rule for `sushi` instead.

City-busz:
I don't know what you mean, so please do it if you know better.

Issue #856, opened 2024-09-18
Problems with splitting sushi


City-busz:
The split between gnome-sushi and sushi-unclassified does not work well:
https://repology.org/project/gnome-sushi/versions
https://repology.org/project/sushi-unclassified/versions

Most of the packages listed on sushi-unclassified should be in gnome-sushi instead.

AMDmi3:
This is as intended. Packages should have valid upstream URLs to be classified properly. What's left in sushi-unclassified provides dead, unrelated or redirecting URLs.

City-busz:
I don't think so. If you check the homepage links on sushi-unclassified, most of them points to gnome-sushi, which shouldn't be listed there:
 https://repology.org/project/sushi-unclassified/information#Homepage_links

AMDmi3:
> What's left in sushi-unclassified provides dead, unrelated or redirecting URLs.

PR #857, opened 2024-09-18
Fix splitting sushi

City-busz:


AMDmi3:
If I was not clear enough,

AMDmi3:
Invalid solution for problem not related to repology. See #856 for explanation.

AMDmi3:
Is another explanation required?

Splitting into `gnome-sushi` by default makes no sense and defies the purpose of splitting, because let another package named `sushi` appear with non-matching (or without) url, it'll go into `gnome-sushi` and break stuff. That's why this PR is invalid and all split rules must fallback to moving packages to `-unclassified`.

We only use valid upstream URLs for classification, not to spend expensive matches on garbage and to encourage maintainers to keep package metadata up to date. As already mentioned, of 13 homepage URLs currently shown in `sushi-unclassifed` neither is invalid:
- 2 are dead
- 3 point or redirect to gnome.org which is not related to the project
- 3 point or redirect to https://wiki.gnome.org/ThreePointOne/Features/FilePreviewing which is not directly related to the project and is also marked for retirement
- 2 point to pages not related to upstream (wikipedia and ubuntugeek)
- 1 point to directory with tarballs which is not a homepage
- 1 points to unrelated project
- 1 redirects to a valid homepage
If you maintain one of these packages, you need to update it to valid upstream URL https://gitlab.gnome.org/GNOME/sushi

City-busz:
> * 3 point or redirect to gnome.org which is not related to the project
> * 3 point or redirect to https://wiki.gnome.org/ThreePointOne/Features/FilePreviewing which is not directly related to the project and is also marked for retirement
> * 1 point to directory with tarballs which is not a homepage
> * 1 redirects to a valid homepage

I think these are valid URLs, just need to match the word 'gnome', not the whole URL to the project repository. See: https://github.com/repology/repology-rules/pull/858

AMDmi3:
This deliberately tolerates invalid urls. Why do you need this?

City-busz:
> This deliberately tolerates invalid urls. Why do you need this?

Because otherwise I can't see the Debian and Ubuntu packages on the gnome-sushi page. Debian sets the homepage of sushi to http://git.gnome.org/browse/sushi, which is the old project URL, and it redirects to the new project URL, so I think it's valid.

AMDmi3:
@jbicha hey, can't we update homepage url for debian/ubuntu gnome-sushi packages?

City-busz:
And openSUSE sets the homepage to https://www.gnome.org, which is actually valid, since sushi is part of GNOME, and https://gitlab.gnome.org/GNOME/sushi is just the project URL, not a homepage.

AMDmi3:
gnome.org is not valid for any specific gnome project. It doesn't have any project specific information.

jbicha:
I can update the Homepage for Debian and Ubuntu to https://gitlab.gnome.org/GNOME/sushi which appears to be the best available page currently. However, this will only take effect for Ubuntu 24.10 and newer and Debian Testing/Unstable. I cannot do a Stable Update for this.

jbicha:
I made the change for Debian Unstable, Debian Testing and Ubuntu 24.10. I do think it looks bad to have so many things at https://repology.org/project/sushi-unclassified/versions that cannot be fixed in the distros themselves. I encourage you to reconsider matching on gnome.org which would be enough to differentiate GNOME Sushi from the old IRC suite.

PR #858, opened 2024-09-18
Fix splitting sushi

City-busz:
Check only for 'gnome' in the URL.


AMDmi3:
As mentioned, not using invalid urls for split

PR #859, opened 2024-09-19
Merge gnome-snapshot with snapshot

City-busz:


AMDmi3:
Like sushi, needs split rule instead

PR #860, opened 2024-09-20
Fix splitting chatty

City-busz:


AMDmi3:
We're not using invalid urls for splitting

PR #862, opened 2024-09-22
Merge libtimezonemap, split nux, and more Chimera Linux rules

EarldridgeJazzedPineda:


850.split-ambiguities/n.yaml
```diff
@@ -141,6 +141,12 @@
 - { name: nut, wwwpart: lanl, setname: nut-neutrino-transport }
 - { name: nut, addflag: unclassified }
 
+- { name: nux, ruleset: openmandriva, setname: nux-toolkit }
```
AMDmi3, line 144: Please remove, we don't split based on ruleset

850.split-ambiguities/n.yaml
```diff
@@ -141,6 +141,12 @@
...
+- { name: nux, wwwpart: launchpad, setname: nux-toolkit }
```
AMDmi3, line 145: Won't it be better to name it `nux-opengl-toolkit`?

200.blacklists/metaports.yaml
```diff
@@ -132,6 +132,8 @@
 - { remove: true,   name: xorg-truetype-fonts }
 - { remove: true,   name: xorg-x11-apps }
 - { remove: true,   name: xorg-x11-utils }
+- { remove: true,   name: xserver-xorg }
```
AMDmi3, line 135: These may remove real packages in other repos, so let's limit these two with `ruleset: chimera`.

AMDmi3 (changes requested):

AMDmi3 (approved):

PR #865, opened 2024-09-28
Add verignore for sgt-puzzles and even more rules for Chimera Linux

EarldridgeJazzedPineda:


800.renames-and-merges/k.yaml
```diff
@@ -31,6 +31,7 @@
 - { setname: kdiff3,                   name: kdiff3-qt }
 - { setname: kdnssd,                   name: kdnssd-framework }
 - { setname: kdsoap,                   name: kd-soap }
+- { setname: kdsoap-ws-discovery-client, name: kdsoap-ws-discover-client }
```
AMDmi3, line 34: Please remove, we don't merge unjustified incorrect naming

AMDmi3 (changes requested):

PR #866, opened 2024-09-30
merge libudis86 and udis86-git to udis86

earboxer:


800.renames-and-merges/u.yaml
```diff
@@ -6,6 +6,7 @@
 - { setname: ubisoft-connect,          name: ubisoftconnect }
 - { setname: uchardet,                 name: libuchardet }
 - { setname: ucsc-genome-browser,      name: [kent, kent-tools] }
+- { setname: udis86,                   name: [libudis86, udis86-git]}
```
AMDmi3, line 9: udis86-git would need separate rule with `weak_devel: true, nolegacy: true`

AMDmi3 (changes requested):

800.renames-and-merges/u.yaml
earboxer: What is `weak_devel` for? I thought `nolegacy` was for packages which can't be marked as outdated -- (not that this will be outdated, since the upstream package is dead), but this isn't an AUR -git package, so it can be outdated.

AMDmi3: `weak_devel` is currently an alias to `ignored`, because -git projects have date or git commit suffix in most cases. `nolegacy` prevents it from being marked as legacy when stable version is higher, as having -git package in that case makes no sense. This pair of flags is used for all -git/-devel/-nighly and alike packages.

PR #867, opened 2024-09-30
f-droid: Add "android:" prefix and merge as needed

Pandapip1:
This will allow for comparing repositories that contain android packages (such as [izzyondroid](https://apt.izzysoft.de/fdroid/), which I will be adding), as well as allowing F-Droid's statistics (particularly https://github.com/repology/repology-updater/issues/1229 and current versions) to be correctly shown


AMDmi3 (changes requested):
I'd retain the current ruleset structure, adding a flag (such as `addflag: android_ok`) instead of removing the package, and then excluding these from prefix rule (e.g. `setname: android:$0, noflag: android_ok`).

Pandapip1:
Not sure I quite get what you meant. Is this closer to what you want?

Pandapip1:
Done!

AMDmi3 (approved):

PR #869, opened 2024-10-20
Merge servarr stack

txtsd:
It's my first time contributing in this repo.

I'm going to use `${servarr}` as a variable to denote any *arr app to explain what this PR aims to achieve.

`${servarr}` and `${servarr}-bin` should be the same version.
`${servarr}-develop` and `${servarr}-develop-bin` should be the same version, and they are development versions.
`${servarr}-nightly` and `${servarr}-nightly-bin` should be the same version, and they are also development versions.

Please let me know if my yaml-fu is correct, or if it needs changing~

### DIFF
--- a/800.renames-and-merges/l.yaml
+++ b/800.renames-and-merges/l.yaml
@@ -84,6 +84,7 @@
 - { setname: librewolf,                name: [librewolf-app, librewolf-browser] }
 - { setname: libtool,                  name: libtool-bin, addflavor: bin }
 - { setname: license-finder,           name: [licensefinder, ruby:license-finder] }
+- { setname: lidarr,                   name: [$0-develop, $0-nightly], weak_devel: true, nolegacy: true }
 - { setname: lightdm,                  name: lightdm-devel, weak_devel: true, nolegacy: true }
 - { setname: lightdm,                  name: lightdm-ubuntu, addflavor: ubuntu }
 - { setname: lightdm,                  namepat: "(liblightdm(?:-qt4|-qt5)*)-devel", addflavor: $1, weak_devel: true, nolegacy: true }
--- a/800.renames-and-merges/p.yaml
+++ b/800.renames-and-merges/p.yaml
@@ -356,6 +356,7 @@
 - { setname: protonmail-bridge,        name: protonmail-bridge-bin, addflavor: bin }
 - { setname: protontricks,             name: "python:protontricks" }
 - { setname: protonvpn-cli,            name: [protonvpn-cli-2, protonvpn-linux-cli] }
+- { setname: prowlarr,                 name: [$0-develop, $0-nightly], weak_devel: true, nolegacy: true }
 - { setname: prusaslicer,              name: [slic3r-prusa, prusa-slicer, slic3r-prusa3d, slic3r-prusa-edition] }
 - { setname: prysk,                    name: python:$0 }
 - { setname: psautohint,               name: "python:psautohint" }
--- a/800.renames-and-merges/b.yaml
+++ b/800.renames-and-merges/b.yaml
@@ -38,6 +38,7 @@
 - { setname: bats,                     name: [bats-core,bash-bats] }
 - { setname: battery-monitor,          name: battery-monitor-devel, weak_devel: true, nolegacy: true }
 - { setname: baudline,                 name: linux-baudline, ruleset: freebsd }
+- { setname: bazarr,                   name: $0-beta, weak_devel: true, nolegacy: true }
 - { setname: bazel,                    name: bazel-clang38, addflavor: true }
 - { setname: bazel,                    namepat: "bazel[0-9.-]+" }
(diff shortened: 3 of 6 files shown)


AMDmi3 (changes requested):
- Please don't put these into separate file, this makes it harder to look for rules. Rulest should be spread alphabetically.
- Please only merge names which are not currently merged. For instance, for `sonarr`, only `sonarr-develop` needs merging, not `sonarr-bin`, `sonarr-develop-bin`, `sonarr-nightly`, or `sonarr-nightly-bin`
- `-develop` and `-nighly` merges need `weak_devel: true, nolegacy: true` instead of `flavor`

txtsd:
Rules can be easily `grep`'d for. I didn't think that would be a problem.

Regardless, I'll make the changes!

txtsd:
Okay, ready for re-review!

AMDmi3 (changes requested):
Like I've mentioned,
- You don't need flavors for nightly/develop. Note that you can also stick nightly and develop into the same rule where applicable.
- Please don't merge non-existing projects. Check project search first, if there's no e.g. `whisparr-nightly-bin`, rules should not mention it.

> EDIT: Are -bin packages automatically classified under their main project? If so, should I remove the -bin for sonarr?

This depends on repository. They are for aur but aren't for gentoo, and that rule is needed for gentoo, so no, please don't remove it.

txtsd:
The nighties and develops should be the same version each respectively though. Is that handled without flavors?

AMDmi3:
> The nighties and develops should be the same version each respectively though. Is that handled without flavors?

Repology does not check version equivalence between packages. develop/develop-bin/nightly/nightly-bin will be handled independently, but the same way, ignored if version higher than release version and outdated if lower.

txtsd:
Okay, ready for review again!

The *-nightly packages on the AUR have been merged into *-nightly-bin, so there aren't any *-nightly, there are only *-nightly-bin now.

txtsd:
Rebased on master

800.renames-and-merges/p.yaml
```diff
@@ -356,6 +356,7 @@
 - { setname: protonmail-bridge,        name: protonmail-bridge-bin, addflavor: bin }
 - { setname: protontricks,             name: "python:protontricks" }
 - { setname: protonvpn-cli,            name: [protonvpn-cli-2, protonvpn-linux-cli] }
+- { setname: prowlarr,                 name: [$0-develop, $0-develop-bin, $0-nightly-bin], weak_devel: true, nolegacy: true }
```
AMDmi3, line 359: https://repology.org/projects/?search=prowlarr
there is no `prowlarr-develop-bin` and no `prowlarr-nighly-bin`

800.renames-and-merges/l.yaml
```diff
@@ -84,6 +84,7 @@
 - { setname: librewolf,                name: [librewolf-app, librewolf-browser] }
 - { setname: libtool,                  name: libtool-bin, addflavor: bin }
 - { setname: license-finder,           name: [licensefinder, ruby:license-finder] }
+- { setname: lidarr,                   name: [$0-develop, $0-develop-bin, $0-nightly-bin], weak_devel: true, nolegacy: true }
```
AMDmi3, line 87: https://repology.org/projects/?search=lidarr
there is no `lidarr-develop-bin` and no `lidarr-nighly-bin`

AMDmi3 (changes requested):
> - Please don't merge non-existing projects

This has not been fixed yet. See comments for examples.

txtsd:
Now it makes sense!

I was thinking in terms of packages, not projects.

I've got the hang of it now for future contributions!

Ready for (hopefully final) review~

txtsd:
Can you also show me how to mark some repositories' packages as develop?

For example in radarr, `Chocolatey`, `Gentoo`, and `LiGurOS develop` package the develop versions of radarr so the actual stable versions are shown as outdated.

AMDmi3 (approved):

AMDmi3:
> Can you also show me how to mark some repositories' packages as develop?

We just [mark](https://github.com/repology/repology-rules/commit/39c5e6ef0258bd3a90a98259e4629eac81abedfa) these repositories as untrusted. Marking version range as devel is possible, but that would require regular manual updating of a rule, we don't want that.

txtsd:
Okay I'll open a new PR for the rest of the packages. Thanks for showing me how!

I'm the maintainer of `bazarr-beta`. I can guarantee it will always be beta.

Also, how often is the website updated?

AMDmi3:
> Also, how often is the website updated?

Once each 4-5 hours.

txtsd:
I see `ruleset: liguros`, but I don't see a way to specifically target `LiGurOS Develop`.

AMDmi3:
We just ban whole repository (which provides unstable version) untrusted, you don't need to specify versions for that.
And you can just use `ruleset: gentoo` which would affect both gentoo and liguros.

PR #872, opened 2024-10-23
Extend merge for ruby-gnome2

FabioLolix:
Merge ruby bindings for `gdk4` and `gtk4` to `ruby-gnome2`

I have adopted most of the ruby bindings for gnome on AUR

The rule could be renamed to `ruby-gnome`  since it cover up to gtk4, the name was likely due to the original website which no longer exist (https://ruby-gnome2.osdn.jp/)

It is now developed at https://github.com/ruby-gnome/ruby-gnome


AMDmi3:
> The rule could be renamed to ruby-gnome since it cover up to gtk4

Makes sense, let's do it. Note that we'll likely need a rule that merges ruby-gnome2 into ruby-gnome.

And you may want to merge `ruby:vtk4` as well.

PR #873, opened 2024-10-24
Fix servarr stack versions

txtsd:
Hopefully this is mostly correct!

Is there a way to specify `LiGurOS Develop`?


AMDmi3:
You don't need to specify liguros at all as gentoo covers it.

txtsd:
Better?
There is no stable version of readarr. It's all develop. I don't suppose I need to do anything in that case?

txtsd:
See: https://github.com/Readarr/Readarr/releases

It's only pre-releases, from the develop branch, and there are nightly builds from the azure CI.

AMDmi3:
Oh. Then I'd consider https://github.com/Readarr/Readarr/releases official

PR #876, opened 2024-10-27
Tweak rule for bazarr

txtsd:


800.renames-and-merges/b.yaml
```diff
@@ -38,7 +38,7 @@
 - { setname: bats,                     name: [bats-core,bash-bats] }
 - { setname: battery-monitor,          name: battery-monitor-devel, weak_devel: true, nolegacy: true }
 - { setname: baudline,                 name: linux-baudline, ruleset: freebsd }
-- { setname: bazarr,                   name: $0-beta, weak_devel: true, nolegacy: true }
+- { setname: bazarr,                   name: $0-beta, devel: true }
```
AMDmi3, line 41: `nolegacy` should stay and you don't need `devel` as long is it's versioned properly (`XbetaY`)

AMDmi3 (changes requested):

AMDmi3:
Timeout

txtsd:
Sorry completely forgot about these

PR #881, opened 2024-11-11
fix: merge Buildpacks `pack` CLI entries

achrinza:
Merges OpenSUSE"s package for the Buildpacks `pack` CLI with the other repos.


800.renames-and-merges/p.yaml
```diff
@@ -7,6 +7,7 @@
 - { setname: pacaur,                   name: pacaur-no-ud, addflavor: true }
 - { setname: pace,                     name: [pace-cli, rust:pace-rs] } # different projects, split in 850
 - { setname: pacemaker,                name: pacemaker1.1 }
+- { setname: pack-buildpack-cli,       name: buildpacks-cli }
```
AMDmi3, line 10: Target name should be `pack`, so all project go through 850 split rule.

Also, please merge `pack-cli` in the same rule, while we're here.

AMDmi3 (changes requested):

PR #883, opened 2024-11-22
Set devel version for Adacore gnatformat project (see repology#713)

jrmarino:


AMDmi3:
It's not packaged anywhere. Why's the rule needed?

jrmarino:
ah, I did the wrong package, I meant to do this one:
https://repology.org/project/prettier-ada/versions

but gnatformat is packaged:
https://repology.org/project/gnatformat/versions

I suggest keeping gnatformat (because w versions are possible) and add prettier-ada.

I can update the PR if you agree.

jrmarino:
done.  sorry about the delay.

PR #886, opened 2024-11-29
Split mp3cat, merge mp3cat-go

TheRealKeto:
There seems to be some ambiguity between two projects with the same name. `mp3cat` was [originally written by Tom Clegg](https://github.com/tomclegg/mp3cat), yet [there exists a different project written by someone else with the same name](https://github.com/dmulholl/mp3cat), only it has a different function and is written in Go.

For Termux, Repology already recognizes the correct name of the package (`mp3cat-go`), so this serves as a change that will split the packages in two, which should also fix issues regarding versions, which also differ (0.5 vs. 4.2.2)

### DIFF
--- a/850.split-ambiguities/m.yaml
+++ b/850.split-ambiguities/m.yaml
@@ -294,6 +294,10 @@
 - { name: monocle, wwwpart: packetstormsecurity, setname: monocle-host-discovery }
 - { name: monocle, addflag: unclassified }
 
+- { name: mp3cat, wwwpart: dmulholl, setname: mp3cat-dmulholl }
+- { name: mp3cat, wwwpart: tomclegg, setname: mp3cat-tomclegg }
+- { name: mp3cat, addflag: unclassified }
+
 - { name: mp3check, wwwpart: icculus, setname: mp3check-icculus }
 
 - { name: mp4ff, wwwpart: Eyevinn, setname: "go:mp4ff" }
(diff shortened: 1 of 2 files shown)


AMDmi3:
You're missing a split rule to assign unique name for another kind of mp3cat. You also need to merge mp3cat-go into mp3cat so all mp3cats go through the same split rules.

TheRealKeto:
To clarify, the other kind of mp3cat rightfully has the name `mp3cat`. This is more to clarify that the package from the AUR with the same name belongs outside that category and should instread be renamed to `mp3cat-go`.

In any case, I've also thought about renaming all packages - Tom Clegg's `mp3cat` -> `mp3cat-tomclegg`; `mp3cat-go` -> `mp3cat-dmulholl`. Let me know which works best for Repology!

AMDmi3:
`mp3cat-tomclegg` and `mp3cat-dmulholl` look good.

850.split-ambiguities/m.yaml
```diff
@@ -294,6 +294,13 @@
 - { name: monocle, wwwpart: packetstormsecurity, setname: monocle-host-discovery }
 - { name: monocle, addflag: unclassified }
 
+- { name: mp3cat, wwwpart: dmulholl, setname: mp3cat-dmulholl }
+- { name: mp3cat, wwwpart: tomclegg, setname: mp3cat-tomclegg }
+- { name: mp3cat, addflag: unclassified }
+
+- { name: mp3cat-go, wwwpart: dmulholl, setname: mp3cat-dmulholl }
```
AMDmi3, line 301: That goes into 800 ruleset, just a single rule is needed.

AMDmi3 (changes requested):

AMDmi3 (approved):

PR #908, opened 2024-12-29
Rename wallpapers to ataraxia-wallpapers

Pandapip1:
Technically not a disambiguation since there's only one package with this name, but it's generic and should be renamed to avoid potential future conflicts. If this would be more appropriate for `800.renames-and-merges`, let me know and I'll move it there.


AMDmi3:
Sorry, we don't usually add anticipatory rules.

PR #913, opened 2025-01-11
Added qt6-qtpbfimageformat to qtpbfimageplugin

tumic0:
Added qt6-qtpbfimageformat as used in OpenSUSE to qtpbfimageplugin

### DIFF
--- a/800.renames-and-merges/q.yaml
+++ b/800.renames-and-merges/q.yaml
@@ -72,7 +72,7 @@
 - { setname: qtkeychain,               name: [qt4-qtkeychain,qt5-qtkeychain,qt5keychain,qtkeychain-qt4,qtkeychain-qt5,qt5-keychain,libqtkeychain], addflavor: true }
 - { setname: qtoctave,                 name: qtoctave-mac }
 - { setname: qtox,                     name: qtox-alpha, weak_devel: true, nolegacy: true }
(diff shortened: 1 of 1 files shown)


AMDmi3:
This needs two rules, one for qt5 and another for qt6 with corresponding `addflavor`'s.

tumic0:
Why two? Because of some technical reasons? Because from the ideological point of view it is the same SW regardless whether it is used in Qt5 or Qt6.

The current state is quiet broken, you have [qtpbfimageplugin](https://repology.org/project/qtpbfimageplugin/versions) which includes Qt5 **and** Qt6 packages and [qt6-qtpbfimageformat](https://repology.org/project/qt6-qtpbfimageformat/versions) that includes the OpenSUSE Qt6 package. Feel free to discard this merge request and fix it the way you want, but at the end there should be either one single `qtpbfimageplugin` entry with both the Qt5 and Qt6 versions (my preference) or two entries - `qt5-qtpbfimageplugin` and `qt6-qtpbfimageplugin` with the correct list of packages.

AMDmi3:
> Why two? Because of some technical reasons?

Yes. With flavors, if some repository contains both qt5- and qt6- packages with different versions, these are treated as different entities, and not a different versions of the same entity, thus older version is classified as outdated instead of legacy.

> should be either one single qtpbfimageplugin entry with both the Qt5 and Qt6 versions

Sure, that's was the plan.

I'll fix it, thanks for submission!

PR #915, opened 2025-01-30
update obsolete rule for dovecot-pigeonhole

jrmarino:
The version scheme changed to match dovecot.
The previous version was 0.5.21.1 and the current version is 2.4.0. Some repos prepended the pigeonhole version with the dovecot version, e.g. 2.3_0.5.21.1, but the rule to mark these incorrect also marks 2.4.0 incorrect.

Update the rule to mark any version with hyphen or underscore incorrect. Add a second rule to mark any version with more than 4 number groups incorrect (all single digits except third group which has optional second digit).

These should continue to mark prepended version incorrect while letting true dovecot-pigeonhole versions through.


jrmarino:
hmm, this didn't seem to work as expected.
image
The versions with hyphens and underscores should have been marked "incorrect" as well as "2.3.0.5.21.1".  I checked the regex with a tool so I'm not sure what's going on here.

AMDmi3:
It works as expected, outdated versions are not subject to ignores.

jrmarino:
oh, okay thanks!

PR #931, opened 2025-02-25
affine: rename affine-whiteboard to affine

xiaoxiangmoe:


xiaoxiangmoe:
@AMDmi3

doodlewind:
As Affine team member, I fully support this rename. It would align better with our product's identity and avoid any confusion. Thanks for considering the change!

xiaoxiangmoe:
@AMDmi3 how about affine-electron or affine-desktop?

AMDmi3:
Same problem, it doesn't give any idea on what the application does. See other split rules.

AMDmi3:
affine-knowledge-base looks good to me

PR #941, opened 2025-03-18
rem: fix split

cho-m:
Not too sure but I think the split from https://github.com/repology/repology-rules/commit/642b111ab16e1cbb4ccb42b591795a9ea0af2548 may have been impacted by rename added in https://github.com/repology/repology-rules/commit/bba5a504e2a14f57e309abbe0eff1cb53386b94c, e.g. https://repology.org/project/rem-kykim/information redirects to librem

Perhaps an alternative could be renaming the split rule to 
```yaml
- { name: librem, wwwpart: kykim, setname: rem-kykim }
```

### DIFF
--- a/850.split-ambiguities/r.yaml
+++ b/850.split-ambiguities/r.yaml
@@ -107,6 +107,8 @@
 - { name: relic, addflag: unclassified }
 
 - { name: rem, wwwpart: kykim, setname: rem-kykim }
+- { name: rem, wwwpart: [baresip,creytiv], setname: librem }
+- { name: rem, addflag: unclassified }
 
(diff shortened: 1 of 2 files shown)


850.split-ambiguities/r.yaml
```diff
@@ -107,6 +107,7 @@
 - { name: relic, addflag: unclassified }
 
 - { name: rem, wwwpart: kykim, setname: rem-kykim }
+- { name: rem, setname: librem }
```
AMDmi3, line 110: There should be a proper wwwpart based split with `unclassified` fallback.

AMDmi3 (changes requested):

PR #942, opened 2025-03-18
brogue: split brogue-ce

cho-m:
The original brogue does document the CE version as newer versions at https://sites.google.com/site/broguegame/home
> For the most current version of Brogue, please look to Brogue: Community Edition

So it doesn't seem any original brogue will ever get a 1.8 and all 1.8+ is part of `brogue-ce` (starting from https://github.com/tmewett/BrogueCE/releases/tag/v1.8)

Wasn't sure if this needed fallback as other usage of `var{gt,ge,lt,le}` in 850 didn't.

### DIFF
--- a/850.split-ambiguities/b.yaml
+++ b/850.split-ambiguities/b.yaml
@@ -285,6 +285,8 @@
 - { name: brightness, wwwpart: Devorlon, setname: $0-ddcutil }
 - { name: brightness, addflag: unclassified }
 
+- { name: brogue, verge: "1.8", setname: brogue-ce }
+
 - { name: broom, wwwpart: a-camarillo, setname: broom-git-branch-cli }
 - { name: broom, wwwpart: nicoulaj, setname: broom-disk-cleaner }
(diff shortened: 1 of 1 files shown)


AMDmi3:
The submitted change is not acceptable. We should decide now whether `brogue` is superseded by `brogue-ce`, or it has any value on its own. In the former case, `brogue-ce` should be merged into `brogue`, verignore rule removed. Otherwise, it should stay the way it is, all repositories packaging `-ce` as `brogue` ignored for providing fake data.

> The original brogue does document the CE version as newer versions

This suggests that it's indeed superseded.

cho-m:
My concern with superseding is the past reports show at least some people want them to be separate: https://repology.org/project/brogue/report

Will close for now and maybe wait to see if AUR and nixpkgs will continue keeping these as separate packages.

AMDmi3:
Merging this, still. There need to be a very good reason to have unmaintained ancient version, which does not even fetch, while there is officially approved maintained successor.

PR #943, opened 2025-03-18
sapling-scm: unignore git hash appendage

cho-m:
Trying to unignore
https://github.com/repology/repology-rules/blob/0bc7e785f817149908cb4f1887dfa099434957a6/899.version-fixes.global.yaml#L24-L26

So newer version is picked from repositories that don't support `+` and use `.` or `-` which get caught by 899 global rule.

https://sapling-scm.com/docs/introduction/installation
> The [latest release](https://github.com/facebook/sapling/releases/latest) is `0.2.20241203-120811+a2174689`.


AMDmi3:
This should check for upstream version pattern, not duplicate the existing ignore pattern.
`.*[._+-][0-9]{6}[._+-][a-f0-9]{8}` should do.
Please also add a comment that it's the offical version pattern.

PR #944, opened 2025-03-18
msieve: mark svn revisions as incorrect

cho-m:
Trying to ignore the svn revisions by macports' "-devel" package and nix. Could just be `[0-9]+` though not sure if there is any risk of upstream doing version "2" release so scoped to 1000+.

Official tags are up to 1.53: https://sourceforge.net/p/msieve/code/HEAD/tree/tags/

With current development for 1.54: https://sourceforge.net/p/msieve/code/HEAD/tree/trunk/Changes

### DIFF
--- a/900.version-fixes/m.yaml
+++ b/900.version-fixes/m.yaml
@@ -259,6 +259,7 @@
 - { name: msgpack,                                                                         noscheme: true } # no way to tell whether it's c or cpp counterpart
 - { name: msgpuck,                     ver: "2.0.10",                ruleset: fedora,      incorrect: true }
 - { name: msgpuck,                                                   ruleset: fedora,      untrusted: true }
+- { name: msieve,                      verpat: "[0-9]{4,}",                                incorrect: true } # macports and nix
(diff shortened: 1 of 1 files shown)


900.version-fixes/m.yaml
```diff
@@ -259,6 +259,7 @@
 - { name: msgpack,                                                                         noscheme: true } # no way to tell whether it's c or cpp counterpart
 - { name: msgpuck,                     ver: "2.0.10",                ruleset: fedora,      incorrect: true }
 - { name: msgpuck,                                                   ruleset: fedora,      untrusted: true }
+- { name: msieve,                      verpat: "[0-9]{4,}",                                snapshot: true } # macports and nix
```
AMDmi3, line 262: That should be `incorrect` instead of `snapshot`. `snapshot` is for e.g. upstream version with date or revision appended, but when it's not based on upstream version it's just incorrect.

AMDmi3 (changes requested):

PR #945, opened 2025-03-18
morpheus: split

cho-m:
Move some of these into existing https://repology.org/project/morpheus-modeling/versions


850.split-ambiguities/m.yaml
```diff
@@ -313,6 +313,9 @@
 - { name: monocle, wwwpart: packetstormsecurity, setname: monocle-host-discovery }
 - { name: monocle, addflag: unclassified }
 
+- { name: morpheus, wwwpart: [morpheus.gitlab.io, morpheus.lab], setname: morpheus-modeling }
```
AMDmi3, line 316: There are like 4 projects, a rule is needed for each. You also need to merge `morpheus-modeling` to `morpheus` so it also passes through this very split rule. This way the project is split reliably even if target names are changed.

AMDmi3 (changes requested):

850.split-ambiguities/m.yaml
cho-m: Updated but wasn't sure on best name for https://github.com/cwenger/Morpheus. May need to be improved.

PR #947, opened 2025-03-19
bomber: update split

cho-m:
Following are merged:
* https://repology.org/project/bomber-go/versions 
* https://repology.org/project/bomber-sbom-security-scanner/versions

Everything unclassified in https://repology.org/project/bomber-unclassified/information looks like `kde-bomber`


AMDmi3:
Merge is good, split is not - we don't split by invalid/outdated/redirecting urls.

PR #956, opened 2025-03-20
keyring: split

cho-m:
Splitting https://repology.org/project/keyring

### DIFF
--- a/850.split-ambiguities/k.yaml
+++ b/850.split-ambiguities/k.yaml
@@ -47,6 +47,9 @@
 - { name: keyfinder, wwwpart: magicaljellybean, setname: keyfinder-windows }
 - { name: keyfinder, addflag: unclassified }
 
+- { name: keyring, wwwpart: ervandew, setname: keyring-ervandew }
+- { name: keyring, addflag: unclassified }
+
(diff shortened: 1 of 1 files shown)


850.split-ambiguities/k.yaml
```diff
@@ -47,6 +47,11 @@
 - { name: keyfinder, wwwpart: magicaljellybean, setname: keyfinder-windows }
 - { name: keyfinder, addflag: unclassified }
 
+- { name: keyring, wwwpart: ataraxialinux.org, setname: ataraxia-keyring }
```
AMDmi3, line 50: Not a valid url, as does not point to the software, and also seems to be dead.

850.split-ambiguities/k.yaml
```diff
@@ -47,6 +47,11 @@
...
+- { name: keyring, wwwpart: ervandew, setname: keyring-ervandew }
+- { name: keyring, wwwpart: jaraco, setname: python:keyring }
```
AMDmi3, line 52: This is unprefixed python module, we don't touch these.

AMDmi3:
Two projects are invalid, only one should remain.

PR #962, opened 2025-03-27
libdwarf: merge in dwarfdump and dwarfutils

cho-m:


AMDmi3:
Don't we instead want to merge it and also probably dwarfdump into libdwarf? These seem to come from the single source, and libdwarf already has similar sink rule.

cho-m:
Didn't notice those. Will update in a bit.

Unrelated to this, is it possible to add lower bound to some of `libdwarf`'s CVEs? The ones with -∞ are showing up for new version scheme - https://repology.org/project/libdwarf/cves?version=0.11.1

AMDmi3:
> Unrelated to this, is it possible to add lower bound to some of `libdwarf`'s CVEs? The ones with -∞ are showing up for new version scheme - https://repology.org/project/libdwarf/cves?version=0.11.1

Not really, but it can (and should) be done by fixing NVD. I suggest to contact them via [nvd@NIST.gov](mailto:nvd@NIST.gov) or https://nvd.nist.gov/contact/contact-form describing the problem and listing affected CVEs.

PR #968, opened 2025-04-28
amethyst: split

otiskujawa:
Splitting https://repology.org/project/amethyst

- https://github.com/ianyh/Amethyst kept original name
- https://codeberg.org/neoninteger/amethyst seems to be text editor so appended `-editor` for now
- https://github.com/Geoxor/Amethyst is also using `amethyst-player` (https://repology.org/project/amethyst-player)


AMDmi3:
> * kept original name

Please don't, each entry should have a new name. `amethyst-wm` in this case, likely.

PR #970, opened 2025-05-04
merge gvisor, gvisor-tap-vsock

vthriller:
The only thing I'm not sure about is sys-apps/gvisor-bin from Exherbo, which should be covered by 100.prefix-suffix.yaml (similar to gvisor-bin from AUR), yet somehow it is not.


AMDmi3:
Sorry, we don't accept rules for `go:*` names, as go modules have no official naming scheme. This project looks like an application though, in which name it shouldn't use module naming scheme at all. I'm adding `*-bin` normalization rule for exherbo though.

PR #976, opened 2025-06-05
Add merge rule for po4a

gemmaro:
Hello,

This merges [`perl:po4a`](https://repology.org/project/perl%3Apo4a/versions) (packaged by Nix) into `po4a`.
See also Nix package [recipe](https://github.com/NixOS/nixpkgs/blob/689683e45a0b42642ff42b421657ca28c17e0553/pkgs/development/perl-modules/Po4a/default.nix#L33), which fetches the identical po4a source.

Thank you!


gemmaro:
I updated it so that it is now written in `p.yaml`.

PR #977, opened 2025-06-07
Renames for 'win32api-{headers,runtime}' cygwin packages

jon-turney:
`win32api-{headers,runtime}` is the same upstream project as `mingw64-{i686,x86_64}-{headers,runtime}`, called
`mingw:{headers,runtime}` in repology, but these are the unflavoured packages, used when compiling Cygwin-hosted executables (rather than cross-compiling Windows executables).


jon-turney:
Completely unsure if this is the right place to put this.

PR #978, opened 2025-06-09
libgedit-amtk, libgedit-tepl: merge

jbicha:
amtk was renamed to libgedit-amtk and tepl was renamed to libgedit-tepl upstream: [git repos](https://gitlab.gnome.org/World/gedit) and [homepage](https://gedit-text-editor.org/technology.html).

Here's my attempt at merging this.

### DIFF
--- a/800.renames-and-merges/l.yaml
+++ b/800.renames-and-merges/l.yaml
@@ -68,6 +68,8 @@
 - { setname: lexicon,                  name: ["python:dns-lexicon", dns-lexicon] }
 - { setname: lgogdownloader,           name: lgogdownloader-qt5, addflavor: true }
 - { setname: liberal-crime-squad,      name: [lcs,liberalcrimesquad] }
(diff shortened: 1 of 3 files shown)


800.renames-and-merges/l.yaml
```diff
@@ -68,6 +68,8 @@
 - { setname: lexicon,                  name: ["python:dns-lexicon", dns-lexicon] }
 - { setname: lgogdownloader,           name: lgogdownloader-qt5, addflavor: true }
 - { setname: liberal-crime-squad,      name: [lcs,liberalcrimesquad] }
+- { setname: libgedit-amtk,            namepat: "(?:lib)?amtk[0-9.-]*" }
```
AMDmi3, line 71: Better not to use regexps where it's not necessary. Just `name: [amtk, libamtk]` is sufficient here.

AMDmi3 (changes requested):
You also need to update version rules in 900* ruleset.

PR #980, opened 2025-06-12
ocaml:*: mark date-stamped snapshots incorrect

momeemt:
These packages were versioned as `YYYYY-MM-DD-*` in past nixpkgs, but are now versioned as `version-unstable-YYYYY-MM-DD` for correct version comparison. OCaml packages that could not list outdate packages correctly due to past versioning have been fixed.

- [x] `make check`


AMDmi3:
This PR has no effect, as all nix versions are already ignored, the problem should be fixed by updater commit.

The only valid change here is for `ocaml:dypgen`, where bad version comes from another repository, but you just need `sink` rule for this date-based version.

PR #984, opened 2025-07-06
Add rename rule simpl-sim → simplc

waldyrious:
The normalized name matches the actual project's name, and the identifier at sourceforge: https://sourceforge.net/projects/simpl/

### DIFF
--- a/800.renames-and-merges/s.yaml
+++ b/800.renames-and-merges/s.yaml
@@ -181,6 +181,7 @@
 - { setname: simage,                   name: libsimage }
 - { setname: simdjson,                 name: lib$0 }
 - { setname: simh,                     name: simh-classic, addflavor: true }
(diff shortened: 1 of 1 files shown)


waldyrious:
Uh... am I missing something? Did the philosophy change on renames? I don't see anything in the documentation (README), nor in the existing rules, that suggests this kind of rule is undesirable. Please share what context I might be missing :)

AMDmi3:
> Is it the case that renames are only done in the service of merging package names, i.e. normalizing outliers to conform to the identifiers used in other repositories?

Yes

> If so, that's not really explicit in the docs, but I'd be happy to provide that clarification.

That would be welcome.

AMDmi3:
We don't try to solve a task of "searching a project by its canonical name" at all.

PR #986, opened 2025-07-13
antares: split; merge -sql-beta into -sql

Rhys-T:
Splitting https://repology.org/project/antares

- https://arescentral.org/antares - appended `-game` since that seems to be how other splits do it
- https://antares-sql.app/ - appended `-sql` to match the existing [`antares-sql`](https://repology.org/project/antares-sql) found in AUR

Also merged https://repology.org/project/antares-sql-beta found on AUR into the latter.

### DIFF
--- a/800.renames-and-merges/a.yaml
+++ b/800.renames-and-merges/a.yaml
@@ -212,6 +212,8 @@
 - { setname: ansible-lint,             name: "python:ansible-lint" }
 - { setname: ansible-runner,           name: "python:ansible-runner" }
 - { setname: ant,                      name: ant-junit5, addflavor: true }
+- { setname: antares,                  name: antares-sql }
+- { setname: antares,                  name: antares-sql-beta, weak_devel: true, nolegacy: true }
 - { setname: anthy,                    name: libanthy }
 - { setname: antidote,                 name: zsh-antidote }
 - { setname: antimicro,                name: antimicro-qt4 }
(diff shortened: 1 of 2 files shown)


800.renames-and-merges/a.yaml
```diff
@@ -212,6 +212,7 @@
 - { setname: ansible-lint,             name: "python:ansible-lint" }
 - { setname: ansible-runner,           name: "python:ansible-runner" }
 - { setname: ant,                      name: ant-junit5, addflavor: true }
+- { setname: antares-sql,              name: antares-sql-beta, weak_devel: true, nolegacy: true }
```
Rhys-T, line 215: I couldn't find any documentation of what `weak_devel` means, but that plus `nolegacy` seemed to be how most of the `-beta` merges work, so I just went with that.

Merging it into `antares-sql` seemed to make sense to me, but it looked like some others merged into the ambiguous name and let the splits sort it back out, even though the name being merged wasn't ambiguous. Should I be doing that here?

Rhys-T: Gotcha. I'll fix that, then.

May I ask what the purpose of merging then splitting again is, as opposed to merging directly into the correct name? I couldn't find that practice documented anywhere…

Rhys-T: Makes sense. Thanks. I'll get it straightened out in just a minute.

Rhys-T: Should I also preemptively merge `antares-game` for the same reason? No repo currently provides it, but there's [an open issue on Nixpkgs](https://github.com/NixOS/nixpkgs/issues/373336) that proposes adding it under that name.

AMDmi3: I avoid adding rules for nonexistent projects.

PR #1000, opened 2025-08-19
eden: split on `wwwpart: eden-emu`

paveloom:
There is an emulator with the same name: https://eden-emu.dev.


AMDmi3:
Thanks and congrats on 1000th issue

PR #1006, opened 2025-08-26
choose-rust: split and merge

commitmaniac:
Merging https://repology.org/project/choose/versions

- https://github.com/theryangeary/choose is the correct homepage and upstream
- `choose` and `choose-rust` are the same package (Homebrew and AUR use choose-rust)
- `rust:choose` is another reference of the same package, from https://crates.io/crates/choose

(`rust:choose` == `choose-rust`)  =/= [`choose`](https://github.com/geier/choose), thus they should be split into `choose` and `choose-rust`


commitmaniac:
In retrospect, this might also need a split rule, since some package distributors (e.g Homebrew) have other packages strictly named `choose`, which refer to a different project.

PR #1023, opened 2025-10-04
emacs: remove and fix some merges

qweered:
Delete old merge film-emacs25
Move some merges to emacs:
Delete some emacs-* merges to uncover errors needs emacs:* wildcard


AMDmi3:
Multiple unrelated changes mixed up in single PR.

AMDmi3:
You cannot remove emacs-* from emacs: merges, as not all repositories are covered by wildcard rules.
Anyway, please split into separate PRs.

qweered:
> You cannot remove emacs-* from emacs: merges, as not all repositories are covered by wildcard rules. Anyway, please split into separate PRs.

I will cover them in next prs, right now it is hard to know what projects are already fully covered by wildcards and what not. Can we merge it for temporary downgrade so it be easier to add remaining rules?

AMDmi3:
I don't see why we need to clean these up at all.

PR #1026, opened 2025-10-05
qt: extend merge

qweered:
https://repology.org/projects/q/?search=-static


AMDmi3:
> Old rules that no repository uses may be mistakenly used to merge newly introduced packages with same names

In most cases that would be desirable

PR #1029, opened 2025-10-05
cleanup repositories and rules

qweered:
removed eol or not complying repositories that also do not have parsers in repology-updater


qweered:
I want to partially revive this pr. Can we establish policy for this? Eg if repo is 5 years dead like ataraxia, no forks existed or planned, no parser exists, why there is no benefit to remove it? Its completely dead code that require at least some maintenance

AMDmi3:
I do not feel like it requires any maintenance. I am sure though that I'll have to spend significant effort on reviewing and testing these changes and I'm not willing to do this ATM.

PR #1056, opened 2025-10-08
emacs:mediawiki: remove rule

qweered:
(cherry picked from commit 968bc7bfb91d58da377bcaa4bf5767cf60257758)


AMDmi3:
This case should be covered by the wildcard rule, so we just need to remove this rule.

PR #1060, opened 2025-11-15
Add monaspace font set to m.yaml

FranklinYu:
https://monaspace.githubnext.com/


AMDmi3:
The last one should probably go to nerd-fonts instead

PR #1072, opened 2025-12-09
qq: update split

chillcicada:


850.split-ambiguities/q.yaml
```diff
@@ -23,9 +23,10 @@
 
 - { name: qputty, wwwpart: qputty-qt5, setname: qputty-qt5 }
 
-- { name: qq, wwwpart: linuxqq, setname: linuxqq }
-- { name: qq, wwwpart: macqq, setname: macqq }
+- { name: qq, wwwpart: [im.qq.com/index,im.qq.com/linuxqq], setname: linuxqq }
+- { name: qq, wwwpart: im.qq.com/macqq, setname: macqq }
```
AMDmi3: What's the point of adding a hostname?

chillcicada: it would be ambiguous if it only uses /index

AMDmi3: Please leave only JFryy line. qq.com related items which do not reference specific os variant cannot be split unless urls are updated in packages.

PR #1074, opened 2025-12-11
sparkle: split, merge sparkle-electron

chillcicada:


800.renames-and-merges/s.yaml
```diff
@@ -309,6 +309,7 @@
 - { setname: spamassassin-ixhash,      name: [spamassassin-ixhash2, spamass-ixhash] }
 - { setname: spandsp,                  name: [libspandsp,libspandsp6], addflavor: lib }
 - { setname: spark-adacore,            name: spark2014 }
+- { setname: sparkle-mihomo,           name: sparkle-electron }
```
AMDmi3: Like I've said, please merge into `sparkle`. This way it would pass through a split rule, and is less likely to end up in separate project if the split rule is changed.

AMDmi3 (changes requested):

PR #1097, opened 2026-01-01
kubernetes-client-c: Split from kubernetes

dg0yt:
Separate source and versioning: https://github.com/kubernetes-client/c/releases.

Affects vcpkg repository.

### DIFF
--- a/850.split-ambiguities/k.yaml
+++ b/850.split-ambiguities/k.yaml
@@ -137,6 +137,8 @@
 - { name: kubelogin, wwwpart: int128, setname: kubectl-oidc-login } # reused name
 - { name: kubelogin, addflag: unclassified }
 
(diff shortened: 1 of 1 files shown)


dg0yt:
So it is "misnamed" now ... I don't think it is helpful. But there is no dialog here anyways.

dg0yt:
This PR was for "garbage in, added value out". What is merged is "garbage in, garbage out". I acknowledge that this pattern has a fair place in certain contexts. But here?

- AFAICS upstream (kubernetes) actively documents (if not endorses) the names chosen by conan and vcpkg.
- There are kubernetes-client packages for haskell and python here. Now the C client is handled differently.
- I'm not convinced that this implementation of the policy has a positive effect on the size of the ruleset, or on what upstream does, or what the repositories do, or on my position to promote reasonable changes upstream or in the repos.

PR #1111, opened 2026-02-02
aseprite: mark fake AUR version

iamanaws:
Fix aseprite versioning in AUR by stripping the trailing pkgrel component (e.g., mapping 1.3.16.1.1 back to upstream 1.3.16.1).


AMDmi3:
We don't fix invalid versions. You can add a pair of incorrect+untrusted rules.

FabioLolix:
Just checked that the AUR maintainer made the change on the same day

* https://aur.archlinux.org/cgit/aur.git/commit/?h=aseprite-bin&id=28119d3dec46591eb58666b56ca1c9e307b61080
* https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=aseprite-bin

It now use

```
pkgver=1.3.17
_pkgver=${pkgver}-1
```

So it will not poison results

PR #1159, opened 2026-04-14
censor: split

mahlzahn:
See https://repology.org/project/censor/information

Disclaimer: I am myself the developer of the document redaction tool Censor.

### DIFF
--- a/850.split-ambiguities/c.yaml
+++ b/850.split-ambiguities/c.yaml
@@ -140,6 +140,10 @@
 - { name: cemu, wwwpart: ce-programming, setname: cemu-ti-emulator }
 - { name: cemu, addflag: unclassified }
 
+- { name: censor, wwwpart: AKotov, setname: censor-scheduled-blocker }
+- { name: censor, wwwpart: codeberg.org/censor/Censor, setname: censor-document-redaction }
+- { name: censor, addflag: unclassified }
+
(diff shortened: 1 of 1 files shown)


850.split-ambiguities/c.yaml
```diff
@@ -140,6 +140,10 @@
 - { name: cemu, wwwpart: ce-programming, setname: cemu-ti-emulator }
 - { name: cemu, addflag: unclassified }
 
+- { name: censor, wwwpart: AKotov, setname: censor-scheduled-blocker }
```
AMDmi3: This does not look very descriptive, wouldn't `-website-blocker` or `-parental-controls` be better?

AMDmi3 (changes requested):

850.split-ambiguities/c.yaml
mahlzahn: I was unsure about it, but I am fine with any of these. @AKotov-dev: Do you have by chance a preferred option among
- `censor-parental-control`
- `censor-website-blocker`

or something else?

mahlzahn: @AKotov-dev: I now changed it to `censor-website-blocker`.

AMDmi3: Thanks! For the record, we can change it later if there are better ideas.

PR #1167, opened 2026-04-22
cachefetch: add `-bin` suffix

ErenayDev:
Adds `-bin` suffix to `cachefetch`s name list. And adds `addflavor: true` setting.


AMDmi3:
This is not needed

https://github.com/repology/repology-rules/blob/0a15afd36d4c9a56f9b211c92579935e1a26a3ef/100.prefix-suffix.yaml#L27

ErenayDev:
Sorry.

PR #1169, opened 2026-05-16
python: extend wildcards

qweered:
1. https://repology.org/projects/?search=-select&inrepo=macports&
2. https://repology.org/projects/?search=python31&inrepo=macports
3. https://repology.org/projects/p/?search=python3&notinrepo=macports

### DIFF
--- a/200.blacklists/metaports.yaml
+++ b/200.blacklists/metaports.yaml
@@ -65,6 +65,8 @@
 - { remove: true,   name: python2,      ruleset: freebsd }
 - { remove: true,   name: python3,      ruleset: freebsd }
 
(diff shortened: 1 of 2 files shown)


AMDmi3:
There are two unrelated changes here. `-select` part is also invalid as it affects unrelated project.

qweered:
If we could match rules on package description (`common files`) there would be much less false positives. Also match on uniqueness: if no other repository has it 99% safe to remove. Will split other changes tomorrow, I thought they are pretty related and you merge PRs with many commits in past

qweered:
Split into #1171, #1172, #1173.

Issue #154, opened 2019-02-08
Remove more metapackages (task-*)


AMDmi3:
`task-*` pattern indicate metapackages, example is `task-xfce`. Should check all similar names and remove them, maybe identify other patterns.

jtojnar:
In nixpkgs we often have packages that combine a package with its plugins:

https://repology.org/projects/?search=-with-plugins
https://repology.org/projects/?search=-with-extensions
https://repology.org/projects/?search=-wrapper&inrepo=nix_unstable – this sometimes occurs in project names so perhaps only uniques should be skipped

AMDmi3:
metapackages should now always mean the same thing - packages which have no own content which only purpose is to group other packages by depending on them. What was previously called _metapackages_ in repology we now call _projects_, and the old term is mostly only used internally.

AMDmi3:
Closing, since known cases (rosa `task-*`, macports `*-select`) are not trivial to fix (corresponding regexp rule would produce false positives), and removing these entries does not provide much value (they do not cause conflicts and do not even affect freshness statistics). Repositories wishing to hide virtual packages in Repology should convey the status explicitly.

PR #1171, opened 2026-05-17
macports: remove -select metaports

qweered:
https://repology.org/projects/?search=-select&inrepo=macports&


AMDmi3:
What is the point of this change?

qweered:
There is an open issue for removing metaports #154. They are not real projects

AMDmi3:
The issue is now closed. These entries do not do any harm (do not even affect the freshness statistics), wasting any effort on removing them has no sense.

PR #1172, opened 2026-05-17
macports: extend python3NN- wildcard

qweered:
https://repology.org/projects/?search=python31&inrepo=macports


AMDmi3:
Aren't these python flavors and not a modules?

qweered:
You right, sorry, I should've really check this

PR #1174, opened 2026-05-17
freebsd: handle py3NNt- (free-threaded) python ports

qweered:
https://repology.org/projects/?search=py313t&inrepo=freebsd

### DIFF
--- a/500.wildcard.yaml
+++ b/500.wildcard.yaml
@@ -245,7 +245,8 @@
 
 # Python
 - { setname: "python:$0", addflag: wconce, noflag: [wconce,not_wildcard,not_python], ruleset: aosc, category: python }
-- { setname: "python:$1", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}-(.*)", ruleset: [freebsd,pkgsrc,ravenports] }
+- { setname: "python:$1", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}-(.*)", ruleset: [pkgsrc,ravenports] }
+- { setname: "python:$2", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py([0-9]{2,3}t?)-(.*)", ruleset: freebsd, addflavor: $1 }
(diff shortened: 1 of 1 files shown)


500.wildcard.yaml
```diff
@@ -246,6 +246,7 @@
 # Python
 - { setname: "python:$0", addflag: wconce, noflag: [wconce,not_wildcard,not_python], ruleset: aosc, category: python }
 - { setname: "python:$1", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}-(.*)", ruleset: [freebsd,pkgsrc,ravenports] }
+- { setname: "python:$1", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}t-(.*)", ruleset: freebsd, addflavor: free-threaded }
```
AMDmi3: Good catch. But let's instead have a single separate rule for FreeBSD.

AMDmi3 (changes requested):

qweered:
Is this ok? I can split freebsd into two lines for better flavor name

500.wildcard.yaml
```diff
@@ -245,12 +245,13 @@
 
 # Python
 - { setname: "python:$0", addflag: wconce, noflag: [wconce,not_wildcard,not_python], ruleset: aosc, category: python }
-- { setname: "python:$1", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}-(.*)", ruleset: [freebsd,pkgsrc,ravenports] }
+- { setname: "python:$1", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}-(.*)", ruleset: [pkgsrc,ravenports] }
+- { setname: "python:$2", addflag: wconce, noflag: [wconce,not_wildcard,not_python], namepat: "py[0-9]{2,3}(t)?-(.*)", ruleset: freebsd, addflavor: $1 }
```
AMDmi3: You should wrap the whole varying part into the capture for flavor, e.g. `([0-9]{2,3}t?)`

AMDmi3 (changes requested):

PR #1175, opened 2026-05-17
python:zope.*: mark more misnamed

qweered:


AMDmi3:
Multiple unrelated changes in a single PR. At the very least, changes to each distinctive ruleset here should be in a separate PR. Also, please do not spawn regexp rules, these have performance overhead.

qweered:
Split into #1183 #1184 #1185

PR #1176, opened 2026-05-18
nix: prefix home-assistant-custom-components with python:

qweered:
also strips github usernames from project names

verification:
nix_no_name: josephabbey-calendar-export  ->  python:calendar-export
nix_name: python:blakeblackshear-frigate  -> python:frigate


AMDmi3:
1. `python:*` is what comes from pypi 
2. We're not fiddling with any url based package naming schemes (like this one or these used for go modules)
3. Adding rules for projects only packaged in a single repository is pointless

> nix_name: python:blakeblackshear-frigate -> python:frigate

You should've checked that this introduces a name conflict

qweered:
Sorry, can we at least do not add prefix to nix_name so they merged to one project?

AMDmi3:
What is the purpose of having these merged into one project?

qweered:
Cause they literally same projects, nix_name currently incorrectly prefixed with python: as you pointed

AMDmi3:
This does not answer the question

qweered:
Directly answering to your question: right now they have inconsistent naming, if python: prefix should be added only to pypi packages than current nix_name naming is incorrect, most projects in this category have * near their names pointing at possible merge, I propose to unify them by removing prefix so all homeassistant projects from all Nixpkgs versions can be viewed in a single page

AMDmi3:
Still does not. What are you trying to achieve apart from merging repology entries?

PR #1187, opened 2026-05-26
python-rclone: merge into python:rclone-python

stiermid:
The [AUR package](https://aur.archlinux.org/packages/python-rclone) is for [rclone-pyhton](https://github.com/Johannes11833/rclone_python) as it has described in `provides` section.


AMDmi3:
This is not correct. AUR package is misnamed: expected package name for PyPI `rclone-python` would be `python-rclone-python`, not `python-rclone`, and the rule won't work as `python-rclone` is `python:rclone` after 500* ruleset.

PR #1190, opened 2026-06-01
labgrid: merge

AiyionPrime:
@AMDmi3 is there anything I can add in the PR to resolve `labgrid-unclassified` introduced in nixos-26.05 as well?

### DIFF
--- a/800.renames-and-merges/l.yaml
+++ b/800.renames-and-merges/l.yaml
@@ -3,6 +3,7 @@
 - { setname: l-smash,                  name: lsmash }
 - { setname: l-smash-works,            name: [vapoursynth:l-smash-works, vapoursynth:lsmashsource], addflavor: vapoursynth }
 - { setname: labelme,                  name: python:$0 }
+- { setname: labgrid,                  name: python:$0 }
 - { setname: labplot,                  name: [labplot-kf5, labplot2] }
 - { setname: labplot,                  name: $0-nightly, weak_devel: true, nolegacy: true }
 - { setname: labymodlauncher,          name: labymod-launcher }


AMDmi3:
`labgrid` project contains misnamed packages, we do not merge these
`-unclassified` will be fixed by https://github.com/repology/repology-updater/commit/9c4623733b57cf1e206058c1f0297e7cc8e5ee10

AiyionPrime:
Ah great, thank you 🙂

AiyionPrime:
@AMDmi3 Congrats on the `-unclassified` fix, that works great :)

How do I spot the `labgrid` group to contain misnamed packages?
Ideally we'd like to see all the packages to be part of one group, right?

AMDmi3:
> Ideally we'd like to see all the packages to be part of one group, right?

Ideally yes, but if we don't have enough information, such as that `labgrid` from some repository is a python module and should be placed in `python:labgrid`, we cannot do anything.

AiyionPrime:
Modifying a package name by putting the (primary?) language of a package into it's name might lead to a new level of confusion, imho.

Looking at labgrids case, I'd suggest it would be more useful for users of repology to find the package of the project under a common name instead of providing them with the hint that it's developed in python?

AMDmi3:
> Modifying a package name by putting the (primary?) language of a package into it's name might lead to a new level of confusion, imho.

This is in fact prevailing pattern, as packages are commonly named e.g. `py3-labgrid`, `py311-labgrid`, `py3.11-labgrid` or  `liblabgrid-python`, and it is inevitable to avoid package name conflicts for similarly named modules for different languages. This holds for **all** supported repositories and **all** programming language module ecosystems except for python modules in debian (and derivatives) and slackware.

> I'd suggest it would be more useful for users of repology to find the package of the project under a common name instead of providing them with the hint that it's developed in python?

The question is what needs to be done for that and what needs to be done is manually maintaining a rule for each python module packaged in debian or slackware. That is not going to happen.

AiyionPrime:
> what needs to be done is manually maintaining a rule for each python module packaged in debian or slackware. That is not going to happen.

I'm not advocating for that, it would indeed be a silly resolution.

The pattern I am currently seeing is: everything coming from pypi is marked with the prefix `python:` and everything that NixOS reports as python package is as well.

Every python package in NixOS that is built as by-name instead of as a python module is not.

I understand the sentiment splitting packages into namespaces to avoid collisions of unrelated packages.

What I do not yet understand is how to handle the splits that were introduced for packages now, is there a plan for that?

AMDmi3:
If you are now asking about python programs (not modules) which are also present in PyPI and (in some repositories) also have python prefix, which is a different case, these we merge manually.

AiyionPrime:
We're getting closer, I feel it 😁

We assumed there was a distinction between packaged python modules and programs;

Labgrid is both, I think;
so it also being a module prevents it's merge?

If so, do you think there's a way to link them in another way?
After all the packages do come from the same source repo and do share the same codebase.
Following the current choice we're missing out on potential notifications, that e.g. pypi has been updated for every Debian related labgrid package.

AMDmi3:
Its readme clearly says it's a library.

AiyionPrime:
It is indeed a library.
And it provides applications.

Like many python projects it provides them under project.scripts:

https://github.com/labgrid-project/labgrid/blob/master/pyproject.toml#L100C1-L106C40

Which leaves me with my previous questions:

> Labgrid is both, I think;
so it also being a module prevents it's merge?
> 
> If so, do you think there's a way to link them in another way?

AMDmi3:
> If so, do you think there's a way to link them in another way?

Well, no - we don't handle and don't want to handle these cases in any way.

AiyionPrime:
Is there documentation or a discussion for this design decision I may read up on?

AMDmi3:
Not really. There are likely related rejected issues, prs and reports. The policies document was in the works, but kind of stalled.

PR #1196, opened 2026-06-23
satysfi: merge

gemmaro:
This merges ocaml:satysfi (in nixpkgs) into satysfi.


AMDmi3:
This only affects outdated nix branches, so is not needed

PR #1197, opened 2026-06-28
tewi: rename python package to tewi-torrent

anlar:
PyPI package was renamed from `tewi-transmission` to `tewi-torrent`, new versions released only to `tewi-torrent`.

Not sure about Repology rules, maybe `tewi` entry should contain both of them?


AMDmi3 (changes requested):
This needs both python packages merged

PR #1201, opened 2026-07-08
fetchit: split

commitmaniac:
Splitting https://repology.org/project/fetchit/versions

- https://github.com/Ruturajn/fetchit is a fetch tool similar to neofetch written in Rust, advertised as supporting Linux only
- https://codeberg.org/nzuum/fetchit is a fastfetch alternative written in C
- https://github.com/containers/fetchit is tool to manage the life cycle and configuration of Podman containers, written in Go

These packages have been split into the following, respectively

- `fetchit-ruturajn`
- `fetchit-nzuum`
- `fetchit-containers`

Not all 3 serve the same purpose, with each maintaining different project structures (Rust vs. C vs. Golang), Thus, they should be split into different entries.

### DIFF
--- a/850.split-ambiguities/f.yaml
+++ b/850.split-ambiguities/f.yaml
@@ -61,6 +61,11 @@
 - { name: fetch, sourceforge: fetch, setname: fetch-lhanson }
 - { name: fetch, addflag: unclassified }
 
+- { name: fetchit, wwwpart: [fetchit.readthedocs.io, containers], setname: fetchit-containers}
+- { name: fetchit, wwwpart: nzuum, setname: fetchit-sysinfo-nzuum }
+- { name: fetchit, wwwpart: ruturajn, setname: fetchit-sysinfo-ruturajn }
+- { name: fetchit, addflag: unclassified }
+
 - { name: fex, wwwpart: [semicomplete,jordansissel], setname: fex-field-extraction }
 - { name: fex, wwwpart: [uni-stuttgart,belwue], setname: fex-file-exchange }
 - { name: fex, wwwpart: fex-emu, setname: fex-emu }


850.split-ambiguities/f.yaml
```diff
@@ -61,6 +61,11 @@
 - { name: fetch, sourceforge: fetch, setname: fetch-lhanson }
 - { name: fetch, addflag: unclassified }
 
+- { name: fetchit, wwwpart: containers, setname: fetchit-containers }
```
AMDmi3: None of existing packages would match this - `fetchit.readthedocs.io`, maybe? But leave `containers` be also as it may be matched by another packages.

850.split-ambiguities/f.yaml
```diff
@@ -61,6 +61,11 @@
...
+- { name: fetchit, wwwpart: nzuum, setname: fetchit-nzuum }
```
AMDmi3: I'd prefer `fetchit-sysinfo-` prefix for this and another one - still not too long, but more distinctive.

AMDmi3 (changes requested):

AMDmi3 (changes requested):
https://github.com/containers/fetchit/ is now missing completely

850.split-ambiguities/f.yaml
commitmaniac: Given recent comments, can you explain this further? I failed to understand how it wouldn't match any packages. Does this line need to be re-introduced?

AMDmi3: Sorry for misunderstanding. What I've meant is that `wwwpart: containers` would not match any existing packages, as nix fetchit package lists https://fetchit.readthedocs.io/, not https://github.com/containers/fetchit as the homepage. So for a proper split you need to match `fetchit.readthedocs.io` for `fetchit-containers` target. But because some future packages for that project are likele to use https://github.com/containers/fetchit as well, it would make sense to match both urls, as in

`{ name: fetchit, wwwpart: [fetchit.readthedocs.io, containers], setname: fetchit-containers }`

AMDmi3:
Thanks!

PR #1202, opened 2026-07-12
Add version pattern for 'ns' in n.yaml

Teko012:
nixpkgs sets ns-3 version to bare minor, e.g. 47 -> ns-3.47


AMDmi3:
nix problem, should be fixed there

PR #1203, opened 2026-07-13
bottom: merge

chillcicada:
btm is the program of bottom


AMDmi3:
I've reverted this as `btm` contains two projects and should be split instead.

PR #1204, opened 2026-07-15
suck: merge

commitmaniac:
Merging https://repology.org/project/suck/related

- `suck` is a program used to grab news from a remote NNTP news server
- `suck-cnews` is the same program, pointing to the same upstream, that uses the [C News](https://www.dinoex.net/software.html) news distributor in order to handle news servers

`suck` == `suck-cnews`, given that they point to the same upstream. Thus, they should be merged.


AMDmi3:
Same reason as #1206

PR #1207, opened 2026-07-19
Many Stardust XR packages have proper versioning now

Pandapip1:
Self-explanatory


AMDmi3:
Fake date versions are still there and will outdated the real ones of noscheme is removed. So you should generally replace each noscheme with an `sink` rule matching dates, like `verpat: "20[0-9]{2}.*", sink: true`, or, which would be more correct, a pair of incorrect (matching dates) and sink (matching dates, but not the ones past the last version, like `verge: "20000000", verlt: "20160315"` for stardust-xr-atmosphere.

AMDmi3:
Timeout

PR #1209, opened 2026-07-28
goose: extend split

ian-moody:
Some Goose packages still reference the project’s previous GitHub URL:

https://github.com/block/goose

That URL redirects to its new location:

https://github.com/aaif-goose/goose

Packages using the previous URL are currently classified as
`goose-unclassified`. This extends the existing split rule to recognize
`block/goose` and classify those packages as `goose-ai-agent`.


AMDmi3:

PR #1210, opened 2026-07-28
raptor: extend split

ian-moody:
Raptor RDF packages use several variants of the upstream homepage, including:

- `http://librdf.org/raptor/`
- `https://www.librdf.org/raptor`
- `https://librdf.org/raptor`

The existing split rule only matches the scheme-specific `https://librdf.org/raptor/` form, leaving the other variants under `raptor-unclassified`.

Broaden the match to `librdf.org/raptor` so these URL variants are classified as `raptor-rdf`. The existing SeqAn and game classifications remain separate.

Validation: `make check` passes across all 243 YAML files.


AMDmi3:

PR #1211, opened 2026-07-28
exo: extend split

ian-moody:
`exo-unclassified` contains packages from multiple unrelated projects whose current homepage URLs are not covered by the existing split rules.

This change:

- maps the unambiguous `libexo` and `xfce4-exo` package names directly to `xfce-exo`;
- broadens the Xfce URL match to `xfce.org`;
- adds `exolabs.net` for the distributed AI runtime;
- adds the current `ndiscovered.com` and `ndiscover.com/exo-2-0` URLs for the Exo font.

The existing catch-all remains for packages without enough identifying metadata.

Validation: `make check` passes across all 243 YAML files.


AMDmi3:

PR #1212, opened 2026-07-28
mac: extend split

ian-moody:
Classify additional Monkey's Audio packages named `mac` using project-specific homepage and source URL identifiers.

This extends the existing `monkeysaudio.com` match with `monkeys-audio`, `mac-port`, and `supermmx`, while leaving unrelated packages named `mac` unclassified.


AMDmi3:

PR #1218, opened 2026-08-14
alot: merge

commitmaniac:
Merging https://repology.org/project/python:alot/related

- `python:alot` is the appropriate name for https://github.com/pazz/alot
- `alot` maintains most packages which point directly to the same upstream as the former; same homepage

`python:alot` == `alot`. Thus, they should be merged accurately.


AMDmi3:
Should be merged to `alot`, it's not a module

PR #1219, opened 2026-08-14
gdown: merge

commitmaniac:
Merging https://repology.org/project/gdown/related

- `gdown` is a wget/curl-like CLI for Google Drive, hosted at https://github.com/wkentaro/gdown
- `python:gdown` contains entries relating to `gdown`, having the same upstream and homepage; it can be used as a module, per their documentation.

`gdown` == `python:gdown`. Thus, they should be merged.


AMDmi3:
Likewise, if it's an utility, it goes to gdown and not python:

PR #1241, opened 2026-09-20
netwatch: separate the Rust TUI from the 1990s netwatch

matthart1983:
Two unrelated programs share the name `netwatch`:

- **netwatch** (https://slctech.org/~mackay/NETWATCH/netwatch.html), a C tool last released as 1.3.1, packaged by Arch, Gentoo, OpenMandriva, Rosa and others.
- **netwatch** (https://github.com/matthart1983/netwatch), a Rust terminal network monitor, crate `netwatch-tui`, currently 0.32.1. It is already tracked as the `netwatch-tui` project via its AUR and nixpkgs packages.

The homebrew-core `netwatch` formula is the Rust one (https://github.com/Homebrew/homebrew-core/blob/master/Formula/n/netwatch.rb — it builds `netwatch-tui` from that GitHub repo), but it is currently grouped with the C tool, where its 0.32.0 is reported as outdated against 1.3.1.

This moves the homebrew entry to `netwatch-tui`, alongside the AUR and nixpkgs packages for the same program.

I am the author of the Rust netwatch.


AMDmi3:
Thanks! This should've been done in another way and in another part of ruleset, and also needed a related merge rule so I've done it myself.

